PR #20, opened 2011-11-19
Added plone.app.portlets to checkouts.cfg.

fulv:
My fix for ticket#11189 patches plone.app.portlets, so we need to check that out, too.

Related pull requests:
https://github.com/plone/plone.app.portlets/pull/3
https://github.com/plone/Products.CMFPlone/pull/5


garbas:
as well cherry picked into 4.1 branch

PR #24, opened 2012-06-15
Nobody uses simplejson

do3cc:
It is only imported in Products.TinyMCE as a fallback if json is not found, which is hard to achieve in Python 2.6 upwards.

Or is there any good reason not to remove it?

### DIFF
--- a/versions.cfg
+++ b/versions.cfg
@@ -34,7 +34,6 @@ mailinglogger = 3.7.0
 python-dateutil = 1.5
 python-openid = 2.2.5
 repoze.xmliter = 0.5
-simplejson = 2.5.0
 WebOb = 1.0.8
 
 # Plone release


asigottech:
AFAIK plone.app.jquerytoolsuses it as well not just TinyMCE.

Asigottech

do3cc:
jquerytools does not use simplejson.
What I did to check this, I checkout coredev ten minutes ago, ran buildout and did a recursive grep through it in the omelette.
I only bring this up because I wanted to deploy sentry logging with raven
https://github.com/do3cc/raven
and this caused problems because the released versions.cfg all pin simplejson to the antique 2.1.x
So this caused pain and a bit of work to find out what has changed, who uses simplejson and could this affect anything. During this I found out that nobody really uses simplejson
OTOH if many extensions use simplejson they might be unhappy if the pinning is removed. Thats why I did not commit directly.

asigottech:
My bad, I was pulling it in with jquerytools when installing geocaching 
on top.
There are a few that do depended on it but not many and they can deal 
with it themselves I would think.

BR

Wyn

do3cc:
If nobody objects until tomorrow I merge it then

asigottech:
+1

domenkozar:
+1

davisagli:
-1. It's still there because simplejson is still used in preference to the stdlib json library in some cases, because it's faster. Please leave it; we shouldn't just remove version pins from the KGS.

PR #25, opened 2012-06-24
Test config cleanup

davisagli:
@tisto, here's an attempt to rearrange the test egg configuration so we don't have to duplicate it. Does this look all right?


tisto:
@davisagli: Can we merge this into 4.3 as well?

davisagli:
Yeah, done.

PR #26, opened 2012-10-13
Generate `requirements.txt` for use with `pip`

witsch:
This adds a buildout part generating a `requirements.txt` file suitable for installing Plone using [pip](http://pypi.python.org/pypi/pip/):

``` bash
$ virtualenv-2.7 --no-site-packages .
$ bin/pip install -r .../requirements.txt
```


witsch:
mee, this should have gone against 4.3, not 4.2… :p

PR #28, opened 2013-01-04
Remove duplicate entry: plone.rfc822

gforcada:
buildouts don't break, but playing with pip and its requirements.txt format and dependencies showed that plone.rfc822 is added twice (with the same version) on our own versions.cfg.

As is within a section called ecosystem, I don't know if that comes from somewhere else, and thus the source of the duplication will still hit us again...


gforcada:
Forgot to say that on 4.3 there is no duplication.

davisagli:
Thanks!

PR #36, opened 2013-02-06
Update resources/theme/sunburst/rules.xml

Thomvl:
Fixed positioning with only portal-column-one active

### DIFF
--- a/resources/theme/sunburst/rules.xml
+++ b/resources/theme/sunburst/rules.xml
@@ -85,7 +85,7 @@
   </rules>
   <rules if-content="//*[@id='portal-column-one'] and not(//*[@id='portal-column-two'])">
     <copy css:theme="#portal-columns">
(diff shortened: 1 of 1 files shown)


garbas:
@vangheem @TH-code i believe you're owners of plip 13355

vangheem:
This shouldn't be in buildout and should now be removed. @TH-code can you do that? This work is being done in a different package right now. @garbas thanks for the heads up.

TH-code:
@vangheem Yes I'll remove it (was already planning to but havent gotten round to it).

TH-code:
@Thomvl This is an old plip branch, I'm removing it. In the meantime I've added the plip-13355 (note the '-') which has a set of startup diazothemes including boilerplate/initializr themes and sunburst and classic. The pull request should nog be necessary with those. But please try them out. The new plip-13355 basically adds diazotheme.startup which is on the collective.

PR #43, opened 2014-02-19
Update zodb remote

gforcada:
ZODB seems to have been moved to GitHub: https://github.com/zopefoundation/ZODB

Note that jenkins jobs will fail (and for that matter all buildout runs) if they have a previous version of ZODB on src/

Seems that mr.developer is not capable of changing a repository from svn to git, so a removal of that folder _before_ running bin/buildout is needed.


gforcada:
Oops, wrong branch!

PR #52, opened 2014-09-22
Change heading to be less specific

jean:
I was looking at the docs wondering "how do I get started"?
I don't want to commit fixes to the core, I just want to trial a bleeding edge Plone, and the original heading was opaque to me.

### DIFF
--- a/docs/intro.rst
+++ b/docs/intro.rst
@@ -1,5 +1,6 @@
-How To Commit Fixes to Plone Core
+Getting started with development
 =================================
+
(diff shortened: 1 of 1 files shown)


jensens:
- your change is on the 4.3 branch only
- the text below the heading may need some chnages too, so it reflects the different targets.

jean:
It's on 4.3 because that is current stable and there's no _master_ branch. What branch should I commit on so that future release branches will get the change?

jensens:
future plone 5.0 is on the 5.0 branch

jean:
https://github.com/plone/buildout.coredev/pull/53 for 5.0 branch

PR #56, opened 2014-11-01
Remove buildout.dumppickedversions extension

khink:
With a modern buildout this gives me the Error: Buildout now includes 'buildout-versions' (and part of the older 'buildout.dumppickedversions').

I think it's best to remove the buildout.dumppickedversions extension.

### DIFF
--- a/core.cfg
+++ b/core.cfg
@@ -28,7 +28,6 @@ find-links += http://dist.plone.org/thirdparty/
 
 extensions =
     mr.developer
-    buildout.dumppickedversions
 
 versions = versions
 


davisagli:
Plone 4.3 still uses buildout 1.x; did you mean to make this pull request for Plone 5?

Also, can we set `show-picked-versions = true` so we still get similar output via the new buildout 2 feature? (https://pypi.python.org/pypi/zc.buildout/2.0.0#easier-reporting-and-managing-of-versions-new-in-buildout-2-0) It's useful information.

khink:
Thanks for answering. I meant to merge with 4.3, that's where i got the error.

Did you mean Plone 4.3 still uses bootstrap 1? Maurits van Rees committed (Mon Feb 11 18:30:31 2013 +0100) a "Fresh http://downloads.buildout.org/1/bootstrap.py" which "Avoids possible problems with zc.buildout 2.0.0". 

I just ran bootstrap with "-v 1.7.1" and then "-v 2.2.1" and both worked.

The problem being that running bootstrap without pinning the zc.buildout version would print the dumppickedversions message. 

Checking that once more... and now bootstrap works fine. I'm wondering how that happened. No updates in the meantime, there were no manual changes to bootstrap.py. I'll try to reproduce later (on a good internet connection).

PR #60, opened 2015-01-24
Small doc fix for virtualenv bootstrap setup

khink:


khink:
@iham I remember you had the same issue with the current doc.

gforcada:
Merged via #70.

PR #63, opened 2015-02-05
Add notes about build notifier

do3cc:
Put Jenkins higher above, because it is not an afterthought.
Reword some phrases to not server as a cheap excuse not to write tests.
@tisto ?


docs/intro.rst
```diff
@@ -202,62 +236,6 @@ actually merge it to the 4.2 release branch. Let's do that now::
   > git merge my-awesome-feature-4.2
 
 
-Committing to Products.CMFPlone
```
tisto, line 205: @do3cc Why did you delete the entire section?

do3cc, line 205: It was in there twice

PR #66, opened 2015-02-10
Remove reviews and implementation notes

gforcada:
Seems that nearly all of this has +3 years.

_And with this I've deleted more lines than I will ever contribute_ :-)

Small note: I kept the folder `plips/implementation_notes` by adding a dummy `.keep` file on it, is this folder still being used?


gforcada:
As nobody complained for two days, I'm merging it :)

vangheem:
if you know the code/change likely better than anyone else, merge now, ask forgiveness later is what I say. ;)

We can always revert if it's a problem.

tisto:
+1

PR #69, opened 2015-02-25
Remove more merged PLIPs

gforcada:


plips/plip13770-portal-tools-removal.cfg
```diff
@@ -1,49 +0,0 @@
-# https://dev.plone.org/ticket/13770
```
tisto, line 1: @gforcada this PLIP is still relevant: https://dev.plone.org/ticket/13770

gforcada:
Ok, removing it then.

PR #82, opened 2015-06-02
Update setuptools and zc.buildout

gforcada:
The first to 8.3 and the later to latest one available (2.3.1).


gforcada:
Testing http://jenkins.plone.org/job/pull-request/205

vangheem:
Should we not set a setuptools version? zc.buildout + setuptools + virtualenv setuptools version can be painful.

gforcada:
@vangheem at least until we use a newer version of a CMF package (I think CMFCore that @mauritsvanrees fixed) we are stuck at 5.8/8.3... anything above that (in setuptools) breaks our builds

gforcada:
@vangheem and at the end pinning versions is what we do with everything, so why setupools, specially now given their relative high development peace, should be different?

vangheem:
Yah, the old reason would be because of the pain and potential issues you get from pinning that setuptools.

gforcada:
@vangheem so it seems that even 8.3 is _too_ new: http://jenkins.plone.org/job/pull-request/205/testReport/ so yes, :+1: for pinning everything :)

vangheem:
man, that's lame.

thet:
@gforcada did you try with the latest setuptools? IIRC, around 8.something, they changed the versioning policy to be PEP440 compliant. later on, they added backwards compatibility for pre 8.something versions.

gforcada:
@thet no, apart from 5.8 and 8.3 which already breaks, I haven't tried anything else so far, let's give it a try though...

tisto:
Testing setuptools 17.0: https://github.com/plone/buildout.coredev/pull/84

tisto:
http://jenkins.plone.org/job/pull-request/208/

tisto:
Had to revert. Sorry for the noise. See https://github.com/plone/Products.CMFPlone/issues/403 for details.

mauritsvanrees:
The problem is that we need a new GenericSetup release (that will be 1.7.6 once released), because that will sort the upgrade steps correctly with both old and new versions of setuptools.

At the end of May, Tres Seaver has moved the CMF repos from svn.zope.org to github.com/zopefoundation (and launchpad as readonly backup that will be synced). The dust needs to settle a bit before a new release. The core CMF buildout needs some love for a failing CI test: http://permalink.gmane.org/gmane.comp.web.zope.cmf/19239

PR #105, opened 2015-08-24
Test plone app intid on Plone 4.3 branch

mauritsvanrees:
I noticed that plone.app.intid is not tested on 4.3 coredev. And 4.3 is using the master branch, which got some changes recently. The changes may be safe, but one of them at least gives a test failure on 4.3. See https://github.com/plone/plone.app.intid/pull/3

So: we should the package. Or was there a specific reason that we are not doing this yet?

### DIFF
--- a/checkouts.cfg
+++ b/checkouts.cfg
@@ -3,6 +3,7 @@ always-checkout = force
 auto-checkout =
     Products.CMFPlone
     plone.app.locales
+    plone.app.intid
 # Test fixes only (or otherwise not-gonna-release changes)
 #    collective.xmltestreport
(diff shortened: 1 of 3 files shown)


mauritsvanrees:
Oh, the first commit, adding p.a.intid to the checkouts, should probably be done on 4.3 anyway, whether we want these tests or not. Oh well, let's see what Jenkins says:
http://jenkins.plone.org/job/pull-request-4.3/78/

pbauer:
I created a new branch 1.1.x for Plone 4. Tests are passing now.

PR #107, opened 2015-09-07
Add plone.api version pinning

jaroel:


gforcada:
:+1: it should be totally safe to merge right?

tisto:
Yep.

PR #114, opened 2015-09-20
Checkout Products.GenericSetup.

mauritsvanrees:
This is to check https://github.com/plone/Products.CMFPlone/issues/1041

### DIFF
--- a/checkouts.cfg
+++ b/checkouts.cfg
@@ -25,6 +25,7 @@ auto-checkout =
     Products.Archetypes
     Products.CMFPlacefulWorkflow
     Products.CMFPlone
+    Products.GenericSetup
     z3c.form
 # test fixes only
     plone.app.iterate


mauritsvanrees:
http://jenkins.plone.org/job/pull-request-5.0/368/

pbauer:
@mauritsvanrees there are still test-failures. An easy way to reproduce:
`./bin/test -s plone.app.contenttypes -t test_content_profile`

mauritsvanrees:
Hopefully fixed with https://github.com/plone/plone.app.contenttypes/commit/325a63c5f7742c22352b2b3b7c60f94133135939
At least then the p.a.contenttypes tests pass for me locally with old and new GS.

New Jenkins run (in the queue): http://jenkins.plone.org/job/pull-request-5.0/370/

mauritsvanrees:
The problem was: plone.app.contenttypes was correctly installed, I can see that in a pdb in that `setUp` method. Before anything is executed of this method, portal.portal_types has Document and various others in it. Then we would until now apply Products.CMFPlone:plone, which also applies several dependencies, and when we check after this, the portal_types tool is almost empty.
With the old (well current 1.7.7) GS, the subsequent applyProfile of our content profile would also apply its dependency profile-plone.app.contenttypes:default and all would have been well again.

In this case I removed the applyProfile of CMFPlone, because this was totally not needed.

You can compare it with this scenario:
- plone.app.contenttypes is installed
- the user removes the Document type from portal_types (manually, or by applying some other profile that does this)
- the user applies the content profile of plone.app.contenttypes.

Old result: GS ruthlessly applies its dependency profile (p.a.c. default) and all is well. Downside: I think most people here agree that reapplying should be avoided if possible.

New result: GS sees its dependency profile is already is installed, and leaves it alone, giving a traceback when the content profile tries to create a Document.

I guess we should accept that this new result can happen. And in the case of failing tests, we need to fix those tests.

pbauer:
I also had to fix a test in Products.CMFPlacefulWorkflow: https://github.com/plone/Products.CMFPlacefulWorkflow/commit/3c7bd2bf7835b1249ccf8c9cb33e0c6e622480f4
@mauritsvanrees we need to add your changes to the plone-documentation (probably also to the upgrade-guide) and I think we should bump the version of Generic Setup to 1.8

PR #115, opened 2015-09-25
Update pin for collective.z3cform.datagridfield and pin zc.queue to 1.3

ale-rt:
collective.z3cform.datagridfield 0.11 is really old
zc.queue > 1.3 triggers ZODB4 and friends resulting in broken buildouts

### DIFF
--- a/versions.cfg
+++ b/versions.cfg
@@ -237,7 +237,7 @@ zope.schema                           = 4.2.2
 # Ecosystem (not officially part of core)
 collective.elephantvocabulary         = 0.2.5
 collective.js.jqueryui                = 1.10.3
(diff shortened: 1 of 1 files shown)


ale-rt:
Probably this can cherry picked also to the 4.3 branch

mauritsvanrees:
They are currently not pinned in 4.3, because they are not used there by default.

ale-rt:
collective.z3cform.datagridfield was pinned to 0.11 until 4.3.5. Then it was removed.
zc.queue (to my knowledge) was never pinned in the 4.3 branch (it is a requirement from singing and dancing dancing).

I have no strong opinion about cherry picking this to the 4.3 branch.

PR #119, opened 2015-09-28
Updated micro versions of decorator, WebOb, future.

mauritsvanrees:
Note that decorator and future also have new minor versions available,
but let's try micro versions first.
WebOb has a newer minor version, but that is still beta.


mauritsvanrees:
http://jenkins.plone.org/job/pull-request-5.0/418/

mauritsvanrees:
Jenkins is green.

PR #122, opened 2015-09-30
Bump mr.developer version

gforcada:
This finally fixes the git-clone-depth option so CI can be vastly
improved.


gforcada:
Running here: http://jenkins.plone.org/job/pull-request-5.0/428/

PR #124, opened 2015-10-04
Remove obsolete cfgs and bump Pillow to 3.0.0

pbauer:


pbauer:
test: http://jenkins.plone.org/job/pull-request-5.0/434/

gforcada:
Seems P.PortalTransforms fails due to a missing dependency on libtiff, I will look into it

gforcada:
Trying again: http://jenkins.plone.org/job/pull-request-5.0/435

mauritsvanrees:
Jenkins says:

```
AssertionError: [II* ]

!=

[II* ]
```

Locally it is very slightly clearer, with a `x04` versus `x03` when comparing the first twenty bytes of expected and actual output:

```
(Pdb) first
'II*\x00\x08\x00\x00\x00 \x00\x00\x01\x04\x00\x01\x00\x00\x00\x18\x00'
(Pdb) second
'II*\x00\x08\x00\x00\x00 \x00\x00\x01\x03\x00\x01\x00\x00\x00\x18\x00'
(Pdb) first == second
False
```

This is with a small Plone logo. Visually it looks the same to me.

jensens:
The part/changed byte is in the image file directory (IFD), see http://www.fileformat.info/format/tiff/corion.htm

looks like this is related to Pillow Changes for 3.0.0
https://github.com/python-pillow/Pillow/blob/master/CHANGES.rst
"Tiff ImageFileDirectory rewrite #1419 [anntzer, wiredfool, homm]"

and looks like this was fixed for 3.1.0 unrelased (not sure, but sound like):
Flatten tiff metadata value SAMPLEFORMAT to initial value, fixes #1466 [wiredfool]

pbauer:
Then let's wait for 3.1.0 or 3.0.1 and merge it when we don't have to fix a test. It's not urgent.

jensens:
digged deeper, the fix for 3.1.0 seems unrelated python-pillow/Pillow#1466

digging further this is the major change python-pillow/Pillow#1419 I suspect as source of our problems.

jensens:
current test failure seems to be a test setup problem, see plone/jenkins.plone.org#141

gforcada:
@jensens not a test setup rather a junit parser error as the error is still there.

Whenever in doubt just look at the full console log: http://jenkins.plone.org/job/pull-request-5.0/695/console

jensens:
well, but how can a pillow update break the test output. strange. maybe the test is crap.

gforcada:
Because the output is binary :-)
El dia 16/01/2016 17:01, "Jens W. Klein" notifications@github.com va
escriure:

jensens:
So I fixed the output in PortalTransforms to be able to analyse the problem. See plone/Products.PortalTransforms#11

The TIFF header (first 8 Bytes) are ok.

In the following (in test output truncated) IFD headers are changed: the byte at position 13 is a `4` instead of a `3`. This byte indicates the value-type (3 is a word, 4 a dword) of tag `0100` (Image width).

Further reading: http://www.fileformat.info/format/tiff/corion.htm http://paulbourke.net/dataformats/tiff/
I do not see any problem here. I'am not sure what the reason behind this is - mabe there are images wider than short integers max are used - but any code reading IFD headers must support this. 

I propose to change PortalTransforms so it 
- before branch away a 2.1.x branch to be used in Plone 4.3,
- bump version on minor level (2.2.0),
- depend on Pillow>=3.0.1
- uses a expected output test file with a 4 in the header.

Opinions?

thet:
+1

mauritsvanrees:
+1

jensens:
done.

PR #131, opened 2015-10-22
Bump plone.protect

gforcada:
Add plone.protect 3.x


gforcada:
Running on jenkins: http://jenkins.plone.org/job/pull-request-4.3/137/

vangheem:
going to be lots of failures I imagine. We might want to mod the tests to run with CSRF protection disabled for now.

gforcada:
@vangheem yes indeed, but they have to be fixed at some point right? :-)

vangheem:
They are fixed in Plone 5 :) just warning you. I wouldn't waste my time on
it....

gforcada:
Yeah, yet another more reason to move to Plone 5 ASAP, but still, then if one installs Plone 4 and sees the security advice to install plone4.csrffixes she will find lots of situations where a confirmation dialog is shown... CMFPlone itself with this PR for example https://github.com/plone/Products.CMFPlone/pull/1147

jensens:
a trial with plone4.csrffixes included in #132, i know that it is still a problem. But we need to support 4.3 for a while and need some solution.

gforcada:
So let's take your branch and keep working there.

PR #133, opened 2015-10-23
Checkout diazo and update related pins.

mauritsvanrees:
The pins are needed, because otherwise buildout already fails, at least the WebOb pin.

There are some test failures, but let's see this on Jenkins, in case it is something wrong on my machine.

See also https://github.com/plone/diazo/issues/51

### DIFF
--- a/checkouts.cfg
+++ b/checkouts.cfg
@@ -2,6 +2,7 @@
 always-checkout = force
 auto-checkout =
     archetypes.referencebrowserwidget
+    diazo
     plone.app.contentrules
     plone.app.dexterity
     plone.app.folder
--- a/tests.cfg
+++ b/tests.cfg
@@ -341,6 +341,7 @@ exclude =
     feedparser
     five.globalrequest
     five.grok
+    future
     grokcore.*
     initgroups
     lxml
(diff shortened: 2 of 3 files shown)


mauritsvanrees:
http://jenkins.plone.org/job/pull-request-4.3/147/

Note: this should not be merged, as there are no changes in diazo. I just want to check what the branch does, and maybe create a new branch 1.0.x to avoid problems.
The updated pins might be fine, although they are not needed with 1.0.x.

mauritsvanrees:
Not bad actually. The `future` package gave test failures, but we can exclude them.
Try again: http://jenkins.plone.org/job/pull-request-4.3/148/

mauritsvanrees:
Next run is probably fine, but Jenkins uses Python 2.7 for the pull requests, and Python 2.6 will fail.

mauritsvanrees:
Jenkins on 2.7 is happy.
And I have fixed the tests on Python 2.6 with commit https://github.com/plone/diazo/commit/b2d90031059eefdce96ddab4b8b6efe866a9ee5d
All green. I am going to go ahead and merge this one.

PR #135, opened 2015-10-28
Added short guidelines for contributing.

mauritsvanrees:
This page should be linked to from CONTRIBUTING.rst in all packages.
See https://github.com/plone/Products.CMFPlone/issues/1179


mauritsvanrees:
Let me just merge this. We can tweak the words later. It is good to have it in the open on docs.plone.org so that we can start linking to it.

gforcada:
@mauritsvanrees late, but anyway: thanks a lot!

Issue #101, opened 2015-07-30
5.0 buildout with "auto-checkout = *" broken


jensens:
because:

First: checkouts of `z3c.formwidget.query` and `Products.ExternalEditor` (svn) are broken
Second::

```
The constraint, 2.13.2, is not consistent with the requirement, 'ExtensionClass>=4.1a1'.
While:
  Installing.
  Getting section instance.
  Initializing section instance.
  Installing recipe plone.recipe.zope2instance.
Error: Bad constraint 2.13.2 ExtensionClass>=4.1a1
```

gforcada:
I spent some time on that since PLOG, but specially the svn thing is really frustrating... my idea was to be able to have **everything** on `src/` so we can run all the tests and thus refactoring can happen.

gforcada:
Seems to be working now on 5.1

PR #136, opened 2015-10-31
Checkout all packages.

mauritsvanrees:
Do NOT merge this. Just a test to see how much (if anything) would break if we did this.

This is for issue #101.

Running buildout at least works. Starting an instance on the foreground works too.
I have started the tests locally and see several test failures coming in, at least ZCatalog raising a TypeError 'None cannot be in an index.'

### DIFF
--- a/checkouts.cfg
+++ b/checkouts.cfg
@@ -1,11 +1,3 @@
 [buildout]
 always-checkout = force
-auto-checkout =
-    mockup
-    plone.app.locales
-    plone.app.upgrade
-    plone.app.widgets
-    plonetheme.barceloneta
-    Products.CMFPlone
-    Products.GenericSetup
-# test-only fixes:
+auto-checkout = *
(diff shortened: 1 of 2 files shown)


mauritsvanrees:
Let's see what happens. http://jenkins.plone.org/job/pull-request-5.0/520/
Note that on 4.3 this is actually working fine (after a few recent fixes).

mauritsvanrees:
Another one with ZCatalog branch 2.13:
http://jenkins.plone.org/job/pull-request-5.0/521/

gforcada:
I have this jenkins job as well: http://jenkins.plone.org/view/All/job/all-checkouts/16/

Running on this branch https://github.com/plone/buildout.coredev/tree/all-sources

My goal is to be able to run all tests from all packages, so at some point we can get a stable - all git checkouts - snapshot from where to start refactorings.

mauritsvanrees:
Ah, right, I saw problems with ZCatalog master so I looked for the nearest branch, which was 2.13. I should have checked the versions. We are at 3.0.2 indeed on Plone 5.

Hm, strange that the first Jenkins job, with ZCatalog master passed. Looks like not all packages were checked out.
Anyway, this pull request is not meant for merging, so I'll close it now.

PR #137, opened 2015-11-04
WIP try ZODB3

gforcada:
**DO NOT MERGE**

This pull request tries the latest ZODB3 versions, which are a major release on all of its subprojects (to bring python 3 compatibility), see https://pypi.python.org/pypi/ZODB3

### DIFF
--- a/versions.cfg
+++ b/versions.cfg
@@ -18,7 +18,9 @@ Sphinx = 1.3.4
 # required for recent z3c.form and chameleon
 zope.pagetemplate = 3.6.3
 # Support for more complex catalog queries
-Products.ZCatalog = 3.0.2
+Products.ZCatalog = 3.1
+# Several tests rely on an old-style pytz release:
+pytz = 2013b
 
 
 # Build tools
@@ -101,6 +103,13 @@ z3c.recipe.sphinxdoc = 1.0.0
 # If you have a dependency somewher on persistent, you must
 # try to get rid of it.
 # persistent = 4.0.6
+
+# ZODB
+ZODB3 = 3.11.0
+ZODB = 4.2.0
+persistent = 4.1.1
+BTrees = 4.1.4
+ZEO = 4.1.0
 
 # Plone release
 Plone                                 = 5.0.2


gforcada:
Jenkins http://jenkins.plone.org/job/pull-request-5.0/528

thet:
wow, that went green! this might mean, we can directly include ZODB >= 4.0 instead, including the dependencies listed in https://github.com/zopefoundation/ZODB3/blob/master/setup.py

gforcada:
@plone/release-team should that get PLIP'ed? I'm unaware of anything else besides tests that should be checked though...

Jenkins (again although green on the first run): http://jenkins.plone.org/job/pull-request-5.0/535

thet:
Upgrading should be checked too

On November 9, 2015 1:06:29 AM GMT+01:00, Gil Forcada Codinachs notifications@github.com wrote:

> @plone/release-team should that get PLIP'ed? I'm unaware of anything
> else besides tests that should be checked though...
> 
> Jenkins (again although green on the first run):
> http://jenkins.plone.org/job/pull-request-5.0/535
> 
> ---
> 
> Reply to this email directly or view it on GitHub:
> https://github.com/plone/buildout.coredev/pull/137#issuecomment-154888302

## 

Sent from my Android device with K-9 Mail. Please excuse my brevity.

tisto:
A PLIP would be good to be able to keep track of this. I guess this upgrade would require a Plone 5.1 release. The PLIP does not need to be long. Just mark this issue as a PLIP and add a few lines regarding the change and the possible risks.

thet:
@gforcada PLIP decisions should be directed to @plone/framework-team 

Since ZODB3 3.11.0 depends on `BTrees >= 4.0.0dev`, you should also depend on `Products.ZCatalog >= 3.1`. This one includes a change, which raises a TypeError when a `None` value is indexed:

> Raise a TypeError when trying to index or lookup `None` in an UnIndex. This is a required change for BTrees 4.0+ compatibility, which prevents objects without a clear ordering definition from being inserted in a tree.

This will result in a little more work, since depending on this ZCatalog version will result in some test failures. We need to make sure, we don't index `None` values in our indexers but a marker value instead, if we need to.

There a related discussion on the [Zope-dev] mailinglist: http://comments.gmane.org/gmane.comp.web.zope.devel/30447
And a related Github issue: https://github.com/zopefoundation/Products.ZCatalog/issues/5

jensens:
great news, yes plip!

bloodbare:
I want it!

gforcada:
@thet thanks for the pointers, and yes, sorry I meant to mention the framework team not the release team :)

With the new set of version pins: http://jenkins.plone.org/job/pull-request-5.0/538

jensens:
I this in Scope of Zope4 already?

gforcada:
@jensens makes sense to bundle it with Zope 4 effort and/or being done afterwards...

PR #138, opened 2015-11-04
Remove txtfilter source

gforcada:
It does not have any release on PyPI,
nothing depends on it and the last real code change was back in 2006.

Anyone has any idea if that code is useful at all?


mauritsvanrees:
When I grep through the code in Plone 4.3 I only see a match in plone.app.controlpanel and wicked.  p.a.controlpanel imports from wicked.txtfilter. Maybe the txtfilter package originated there.
Seems fine to remove from our sources.cfg.

PR #147, opened 2015-11-24
fix: update feedparser to 5.2.1

Gagaro:
feedparser 5.0.1 is almost five years old. 5.2.1 was released last July.

Please see a full list of changes there: https://github.com/kurtmckee/feedparser/blob/develop/NEWS

This package is only used by RSS portlet, it seems to works the same with the updated version.


gforcada:
Tests running: http://jenkins.plone.org/job/pull-request-5.0/592

PR #151, opened 2015-12-14
fix links and naming

thet:
- fix links
- change links from docs.plone.org to :doc: references. this makes all those references only work on docs.plone.org. i guess, that's ok...
- change different styles of how github was written to the official naming: GitHub. see: https://github.com/plone/documentation/issues/475

### DIFF
--- a/docs/documentation.rst
+++ b/docs/documentation.rst
@@ -15,7 +15,7 @@ others are included:
 * :doc:`/adapt-and-extend/theming/index`
 * :doc:`/develop/index`
 
-The documentation is hosted on `github <https://github.com/plone/documentation>`_ and information how to contribute can be found at :doc:`/about/contributing`.
+The documentation is hosted on `GitHub <https://github.com/plone/documentation>`_ and information how to contribute can be found at :doc:`/about/contributing`.
 
 Documenting a package
 =====================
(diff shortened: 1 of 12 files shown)


thet:
@polyester @svx plz rvw!

docs/documentation.rst
```diff
@@ -10,13 +10,12 @@ Documentation of Plone
 The comprehensive resource for Plone documentation is http://docs.plone.org/. The following sections among 
 others are included: 
 
-* `User Manual <http://docs.plone.org/working-with-content/index.html>`_
-* `Installing Plone <http://docs.plone.org/manage/installing/index.html>`_
-* `Theme Reference <http://docs.plone.org/adapt-and-extend/theming/index.html>`_
-* `Developer Manual <http://docs.plone.org/develop/index.html>`_
+* :doc:`/working-with-content/index`
```
polyester: please use explicit `name <location>` syntax, as sphinx does not always pick up the title right, and it can be awkward if somebody changes the title. So here: 

```
:doc:`User Manual </working-with-content/index>`
```

thet: done

polyester:
looks good to me

PR #72, opened 2015-03-22
Try latest zope.testrunner

gforcada:

PR #154, opened 2016-01-03
Revert "Try latest zope.testrunner"

gforcada:
Reverts plone/buildout.coredev#72


gforcada:
There's something wrong on jenkins pull request jobs for 5.0 as it failed here: http://jenkins.plone.org/job/plone-5.0-python-2.7-at/3698/ but not here: http://jenkins.plone.org/job/pull-request-5.0/661/ (ok, yes it also failed here, but not the same errors...)

PR #156, opened 2016-01-12
update to latest pytz 2015.7

jensens:
try what happens when updating to recent pytz, there is a comment left saying:

```
# Several tests rely on an old-style pytz release
```


jensens:
branch totally broken, I close this and try again

PR #159, opened 2016-01-14
update tools for test and build to most/more recent versions

jensens:


versions.cfg
```diff
@@ -79,8 +81,38 @@ simplejson = 2.5.2
...
+distribute = 0.7.3
+
+# this is mentioned in zope.toolkit and outdated there - not sure if we need i
+# it at all
+z3c.recipe.sphinxdoc = 1.0.0
+
+# argparse is only needed for Python <2.7 but as zope toolkit 
+# pin an outdaed version be safe to pint to latest
+argparse = 1.4.0 
+argcomplete = 1.0.0
```
gforcada: duplicated of lines 87:90 :-) and if it's only needed for python 2.6 or lower it could be removed all together as Plone 5.0 is only supposed to be used with 2.7

versions.cfg
```diff
@@ -79,8 +81,38 @@ simplejson = 2.5.2
...
+
+# we dont use distribute (outdated fork of setuptools), buts its 
+# pinned in zope tookit. so update to latest legacy package, which is
+# a simple compatibility layer that installs Setuptools 0.7+
+distribute = 0.7.3
+
+# this is mentioned in zope.toolkit and outdated there - not sure if we need i
+# it at all
+z3c.recipe.sphinxdoc = 1.0.0
```
gforcada: duplicated

versions.cfg
```diff
@@ -79,8 +81,38 @@ simplejson = 2.5.2
...
+
+# argparse is only needed for Python <2.7 but as zope toolkit 
+# pin an outdaed version be safe to pint to latest
+argparse = 1.4.0 
+argcomplete = 1.0.0
+
+# we dont use distribute (outdated fork of setuptools), buts its 
+# pinned in zope tookit. so update to latest legacy package, which is
+# a simple compatibility layer that installs Setuptools 0.7+
+distribute = 0.7.3
```
gforcada: dup

versions.cfg
jensens: dups, ops.
instead of a remove i would unset unused if it was inherited from somewhere, because otherwise buildouts inheriting from it may have bad side effects.

jensens:
cleaned up and tests are green.

PR #161, opened 2016-01-22
change repo url from git to https

hackerkid:
Using git instead of https for `git clone` creates [problem](http://stackoverflow.com/questions/5860888/git-through-proxy) in systems which are using a proxy server for server for accessing the Internet. 

> mr.developer: 
> mr.developer: 
> mr.developer: Cloned 'five.intid' with git.
> mr.developer: git cloning of 'five.intid' failed.
> mr.developer: fatal: unable to connect to github.com:
> mr.developer: github.com[0: 192.30.252.131]: errno=Connection refused
> mr.developer: 
> mr.developer: 
> mr.developer: Cloned 'mockup' with git.
> mr.developer: git cloning of 'mockup' failed.
> mr.developer: fatal: unable to connect to github.com:
> mr.developer: github.com[0: 192.30.252.129]: errno=Connection refused
> mr.developer: 
> mr.developer: 
> mr.developer: Cloned 'Products.PortalTransforms' with git.
> mr.developer: git cloning of 'Products.PortalTransforms' failed.
> mr.developer: fatal: unable to connect to github.com:
> mr.developer: github.com[0: 192.30.252.129]: errno=Connection refused
> mr.developer: 
> mr.developer: 
> mr.developer: Cloned 'archetypes.referencebrowserwidget' with git.
> mr.developer: git cloning of 'archetypes.referencebrowserwidget' failed.
> mr.developer: fatal: unable to connect to github.com:
> mr.developer: github.com[0: 192.30.252.131]: errno=Connection refused
> mr.developer: 
> mr.developer: 
> mr.developer: There have been errors, see messages above.

The error is fixed by changing the git to https in the repo urls. Other fix is to force git to use https instead of git

`git config --global url.https://github.com/.insteadOf git://github.com/`

### DIFF
--- a/sources.cfg
+++ b/sources.cfg
@@ -5,15 +5,15 @@
 ########################
 
 # Plone
-plone = git://github.com/plone
+plone = https://github.com/plone
 plone_push = git@github.com:plone
 
 # Collective
-collective = git://github.com/collective
+collective = https://github.com/collective
 collective_push = git@github.com:collective
 
 # Zope
-zope = git://github.com/zopefoundation
+zope = https://github.com/zopefoundation
 zope_push = git@github.com:zopefoundation
 
 


jensens:
+1, but I would like to ask for more opinions from @plone/framework-team and @plone/release-team

mauritsvanrees:
-1. Technically I don't mind, but when we change this, every core developer must throw away his src directory (or do a git command for each repo to switch the url). If I just change it in `sources.cfg`, I get errors like this:

```
mr.developer: Can't update package 'archetypes.referencebrowserwidget' 
because its URL doesn't match.
```

Or are there technical reasons why https is better?

@hackerkid I would say the `git config --global` that you mention yourself  is best here. One command and you are done.

Alternatively, you can create a `local.cfg` in the coredev buildout and use `bin/buildout -c local.cfg`. Contents can be like this:

```
[buildout]
extends = buildout.cfg

[remotes]
plone = https://github.com/plone
collective = https://github.com/collective
zope = https://github.com/zopefoundation
```

jensens:
@mauritsvanrees I agree its annoying to change each repositories .git/config file - but even Github recommends usage of https these days. I personally like the ssh way more, but for the vast majority and those working behind proxys the https way is much more convinient.

see also https://help.github.com/articles/which-remote-url-should-i-use/

We could provide a shell script iterating over all src/*/.git/config files changing the `[remote "origin"]` `url=` entry.

davisagli:
In addition to @mauritsvanrees' comment, my objection is that https it makes it harder to push changes, since you have to enter your password instead of using an ssh key. However, this problem can be worked around with the following in ~/.gitconfig to make it automatically switch to ssh for pushing:

```
[url "git@github.com:"]
    pushInsteadOf = https://github.com/
    pushInsteadOf = git://github.com/
```

mauritsvanrees:
Well, this pull request is only about changing the fetch urls. The push urls remain as they were.

I am a bit worried about this remark in the document @jensens points to:
"When you git clone, git fetch, git pull, or git push to a remote repository using HTTPS URLs on the command line, you'll be asked for your GitHub username and password."
This seems to imply that mr.developer would ask us dozens of times for a username and password. I don't see this though. A `git clone https://github.com/plone/plone.openid.git` works fine and gives me no prompt. When I try to push, I do get a prompt, but we will push elsewhere in coredev.
It could be that the prompt while fetching or pulling is only when a repository is private, which is not relevant for us.

When github itself recommends https, it may be the better choice.

jensens:
Yes, this PR is not enough. 
But we need to discuss how we can better support people behind firewalls and proxys. I know from customers this is a problem.
Btw.: There are different credential managers for git, so you do not need to enter the password all the time.

jensens:
further reading
git-credential-cache - Helper to temporarily store passwords in memory 
https://git-scm.com/docs/git-credential-cache

Works fine here (I have a customers with gitlab and https only).

hackerkid:
@mauritsvanrees 

> When you git clone, git fetch, git pull, or git push to a remote repository using HTTPS URLs on the command line, you'll be asked for your GitHub username and password

Github ask for credentials while cloning only if  `two-factor authentication` is enabled. 

I am looking forward do GSOC with Plone. I would be happy to spend my time on migrating all the urls from git to https. Should I close this pull request and send another one (covering push urls)?

mauritsvanrees:
I do have two-factor authentication enabled. I also have this in my `~/.gitconfig`:

```
[credential]
    helper = osxkeychain
```

Commenting it out yesterday I did not have to enter a password, but maybe git had remembered it from before. Or indeed because these are all public repositories.

I have the impression we do not want to change the push urls at all.

I wonder what Jenkins thinks of the current request, let's see:
http://jenkins.plone.org/job/pull-request-5.0/742/

jensens:
ok, agreed, andwe have indeed other more important problems to solve. 

what about instead an additional cfg "httpscheckouts.cfg" in coredev, one behind firewalls can easily use with "bin/buildout - c httpscheckouts.cfg"?

PR #167, opened 2016-01-26
move grok to ecosystem.cfg

thet:
Our tests install the whole Grok stack via the test-dependency `grokcore.security [test]`.
I moved this and other references to Grok into ecosystem.cfg.

Also moved plone.directives.\* source definitions to ecosystem.cfg, since they are also not part of the Plone core. The reason is, that when checking out all the source definitions, the testrunner picks them up and tries to test, but the runner can't then find the dependency.

/cc @esteele @gforcada


jensens:
+1

gforcada:
I'm +1 on moving grok to ecosystem.cfg but I'm +0 on moving plone.directives.\* source definitions.

A full checkout is not something you want to do every day, but check out those packages is maybe something that will happen more often.

Could the test runner be fixed so that it can find them?

thet:
Updated!
Now the plone.directives sources are kept in sources.cfg

ecosystem.cfg
```diff
@@ -55,6 +56,11 @@ Dexterity-ecosystem =
     plone.formwidget.autocomplete
     plone.formwidget.contenttree
     z3c.relationfield
+Grok =
```
gforcada, line 59: @thet I think this line is not needed, although it can be fixed at any time later on. Am I right @tisto ?

esteele:
Agreed.

PR #169, opened 2016-01-27
Autocheckout p.formwidget.namedfile

staeff:


vincentfretin:
You can do it without a PR. :)

PR #175, opened 2016-02-10
fix buildout restart loop by syncing setuptools version

gyst:
In commit e5e61899 the setuptools version was bumped, without upping the corresponding pin in requirements.txt. This resulted in an endless buildout restart loop.

This fix restores the version synchronicity between versions.cfg and requirements.txt. And documents the need to do so in versions.cfg.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,2 +1,2 @@
-setuptools==19.4
+setuptools==19.6.2
 zc.buildout==2.5.0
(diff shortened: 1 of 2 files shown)


jensens:
Thats a pita with this loops, I had this recently too and currently I have no idea why this happens.

gyst:
@jensens it happens when the pip-installed versions of setuptools and zc.buildout is out of sync with the versions that buildout itself is pinned to. Buildout detects that, upgrades the packages, then restarts itself. Repeat. Repeat.

PR #178, opened 2016-02-20
Move plone.folder to plone_app_testing group

gforcada:
Give it a try now that the error seem to be gone.


gforcada:
So it seems there is more isolation problems... closing by now.

PR #179, opened 2016-02-20
Test isolation problems

gforcada:
plone.testing 5.0 raises an error if a test in an integration test tries to make a commit, which would lead to test isolation problems.

Check how many packages need to be fixed, since quite a few were already fixed.

Not sure if we want to bump plone.testing version in versions.cfg though, as that could lead to client buildout breaking with a minor update... OTOH if we never push it in nobody will be able to fix their own isolation problems either... comments? @plone/framework-team @plone/release-team


tisto:
@gforcada that's actually quite good news! The test isolations problems are just in p.a.linkintegrity and one test in plone.app.layout. 

@pbauer you refactored the p.a.linkintegrity tests lately, right? Would you mind having a look?

gforcada:
There were only errors on [plone.app.linkintegrity](https://github.com/plone/plone.app.linkintegrity/pull/37) and [plone.app.layout](https://github.com/plone/plone.app.layout/pull/80).

Pull requests have been created and jenkins is furiously running them, if both are ok, we can run once more this pull request and see if everything is solved already!

gforcada:
Both pull requests passed and are merged, running jenkins again here to see if everything has been fixed :-)

gforcada:
So there are still some errors (although jenkins fails to show them): http://jenkins.plone.org/job/pull-request-5.0/1010/console

thet:
@gforcada 
In the plone-zope4.cfg PLIP buildout, where we depend also on latest plone.testing, we get these errors reported: http://jenkins.plone.org/view/PLIPs/job/plip-zope4/30/testReport/
68 failures, but -38 after your last isolation fixes.

gforcada:
Closing in favor of #194

PR #181, opened 2016-02-26
Use plone.protect 3.x.

mauritsvanrees:
We currently do not want this in 4.3, but I wonder how many test failures are left after our previous fixes.

DO NOT MERGE.


mauritsvanrees:
87 failures and 14 robot failures.
I'll attach the console log from Jenkins for good measure.
[jenkins-test-failures.txt](https://github.com/plone/buildout.coredev/files/148861/jenkins-test-failures.txt)
These are failures that may pop up currently for users on Plone 4.3.7 with plone4.csrffixes. So they may be worth fixing individually. But likely the effect is that users get the confirmation page, which may be annoying (and needed) but does not break the site.
I am closing this one.

PR #182, opened 2016-02-27
Added documentation about special packages.

mauritsvanrees:
Some packages may be releases by various people.
Others should be done by specific people.
Let's list them here, as I am sometimes wondering if a package needs any special care.

@esteele Can you check this one?
@vangheem I think you know this too.

I think `mockup` may need special handling. If that is true, we should document it here (adding a link is fine).

I was wondering if `plonetheme.barceloneta` and `plone.app.widgets` need special handling, but I see I have made releases of those in the past, and did not do anything special and no one complained. :-)

### DIFF
--- a/docs/release.rst
+++ b/docs/release.rst
@@ -57,9 +57,6 @@ people as they know what they are doing.  These are:
 ``plone.app.locales``:
   Please leave this to the i18n team lead, Vincent Fretin.
 
-``mockup``:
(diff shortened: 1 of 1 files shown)


vangheem:
There actually really isn't anything special about mockup. Releases are made and the tags are auto-added into the bower distribution.

mauritsvanrees:
Okay, I have removed the note about mockup. If the fullrelease script in the coredev buildout is enough, then that's fine. I'll merge this doc change.

PR #183, opened 2016-03-01
Isolation problems fixed

gforcada:
Now that quite some test isolation problems have been fixed, **maybe** these two packages can get back to where they belong.

### DIFF
--- a/tests.cfg
+++ b/tests.cfg
@@ -306,6 +306,7 @@ plone_app_testing =
     plone.contentrules
     plone.dexterity
     plone.event
+    plone.folder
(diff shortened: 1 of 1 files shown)


thet:
@gforcada I'm curious - are there any other test isolation problems left? Does all go green with plone.testing 5.0 in place?

gforcada:
Oh yes, there are plenty of them, LOTS of them actually, that's why tests.cfg is so complex, just take a look at it.

One step at a time though we can reduce that... by now unfortunately my focus is on jenkins and python 3.5 to be able to test the work done at Barcelona sprint.

PR #192, opened 2016-03-08
Rephrase header

svx:
This is needed for https://github.com/plone/documentation/pull/483
- Rephrase header
- Adjust rst to make it bold


svx:
This also should go into 5.0.3 of course !

PR #193, opened 2016-03-08
Rephrase header

svx:
This is in order to fix https://github.com/plone/documentation/pull/483
- Rephrase header
- Change rst syntax to make it bold

### DIFF
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -1,21 +1,20 @@
 .. -*- coding: utf-8 -*-
 
-====================================
-Process for Plone core's development
-====================================
+=========================
+Developing for Plone Core
+=========================
 
 Introduction
 ============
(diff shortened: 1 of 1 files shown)


docs/index.rst
```diff
@@ -1,21 +1,20 @@
...
 
 Introduction
 ============
 This documentation describes the process for developing Plone.
 It is primarily a technical resource for setting up your core developer buildout,
-fixing bugs,
-and writing plips.
+fixing bugs and writing plips.
 
 STOP!
```
gforcada, line 13: @svx _sorry for hijacking your changes_

Can we, pretty please, rephrase this header as well? There's nothing more discouraging than yelling "STOP!"...

Something like _A word of warning_ would be probably less discouraging (other suggestions welcome).

PR #200, opened 2016-03-31
resolve conflicts with our requirement ``ExtensionClass = 4.1.2`` fro…

thet:
…m versions.cfg


jensens:
how does it relates to the releases of this pkgs we're pinning to?

thet:
We have a version pin of `Products.BTreeFolder2 = 2.14.0`. The branch change in sources.cfg is OK for this one. The pin for Products.MailHost comes from http://dist.plone.org/versions/zope-2-13-24-versions.cfg and is `Products.MailHost = 2.13.2`. This version depends on `ExtensionClass < 4.0dev`: https://github.com/zopefoundation/Products.MailHost/blob/2.13/setup.py
Using the latest but unreleased Products.MailHost 2.14 line should be safe.

If we depend on P.MailHost 2.13.2 via Zope2 AND `ExtensionClass = 4.1.2` via our own versions.cfg, I wonder how this could have been build successfully.

thet:
I have to say, I only got these dependency problems when doing a full source checkout with `auto-checkout = *`.

PR #203, opened 2016-04-25
Newer zope.keyreference + deps

jensens:
Introduce here a newer `zope.keyreference` plus its dependencies `zdaemon` and `transaction`. According to their changelogs this is without problems.
Also pinned persistent. This was fetched by coredev and was not pinned.

### DIFF
--- a/versions.cfg
+++ b/versions.cfg
@@ -22,6 +22,8 @@ Sphinx = 1.3.4
 zope.pagetemplate = 3.6.3
 # Support for more complex catalog queries
 Products.ZCatalog = 3.0.2
+# pin here, because not pinned anywhere else
+persistent = 4.1.1
 
 #############
 # Build Tools
--- a/versions.cfg
+++ b/versions.cfg
@@ -13,7 +13,10 @@ Acquisition = 4.2.2
 # More memory efficient version, Trac #13101
 DateTime = 4.0.1
 Products.BTreeFolder2 = 2.14.0
-# Override until ztk is updated
+# Overrides until ztk is updated
+zdaemon = 4.1.0
+transaction = 1.4.4
+zope.keyreference = 4.1.0
 Sphinx = 1.3.4
 # required for recent z3c.form and chameleon
 zope.pagetemplate = 3.6.3
--- a/versions.cfg
+++ b/versions.cfg
@@ -22,8 +22,6 @@ Sphinx = 1.3.4
 zope.pagetemplate = 3.6.3
 # Support for more complex catalog queries
 Products.ZCatalog = 3.0.2
-# pin here, because not pinned anywhere else
-persistent = 4.1.1
 
 #############
 # Build Tools


gforcada:
Seeing the branch name, shouldn't that be part of p.a.async itself rather than on coredev?

jensens:
this are core modules, upgrade must happen anyway now or later.

btw, i give up on p.a.async anyway - but would like to see this merged, and more updates on such outdated ztk deps.

mauritsvanrees:
I am a bit worried about the pin of `persistent`, given the comments in
https://github.com/plone/buildout.coredev/blob/1c4d0b6363e07d948cdab41bd0a9dfb10d91fe01/versions.cfg#L116-L119
It says:

> Persistent is not compatible with ZODB 3.10 which we are using
> If you have a dependency somewher on persistent, you must
> try to get rid of it.

IIRC, last time I looked at this, `persistent` was dragged in as a test dependency, which might be okay. I see it in `bin/test` but not in `bin/instance`.

jensens:
@mauritsvanrees valid point. But my local plain coredev 5.1 buildout always ends with a message:

```
Versions had to be automatically picked.
[versions]
persistent = 4.1.1
```

So we must have pulled in this dependency somehwere - accidentially?

mauritsvanrees:
z3c.form has it in its `[test]` dependency.

jensens:
IMO we should pin it in `tests.cfg` `[versions]` section to be explicit?

jensens:
removed persisten pin here, I'll create a separate PR

jensens:
persistent PR plone/buildout.coredev#204

gforcada:
@plone/release-team given that ztk KGS is already on dist.plone.org, i.e. under our control, shouldn't we add all these 4 versions.cfg here in buildout.coredev? So we can get rid of these overrides altogether.

mauritsvanrees:
We control it only because there is no obvious stable other location to point to.
But it is still being developed by the zope community:
https://github.com/zopefoundation/Zope/blob/master/ztk-versions.cfg

I guess we could point to a github url like this, but dist.plone.org seems nicer:
https://raw.githubusercontent.com/zopefoundation/Zope/2.13.24/ztk-versions.cfg

mauritsvanrees:
Wait a minute, I now see this:

```
Versions had to be automatically picked.
[versions]
BTrees = 4.2.0
zodbpickle = 0.6.0

# Required by:
# zope.keyreference==4.1.0
ZODB = 4.2.0
```

So then we have two ZODB versions in coredev:

```
$ grep ZODB bin/instance 
  '/Users/maurits/shared-eggs/ZODB3-3.10.5-py2.7-macosx-10.4-x86_64.egg',
  '/Users/maurits/shared-eggs/ZODB-4.2.0-py2.7.egg',
```

Are we sure this is sane?

gforcada:
Doesn't feel right, specially if it was not like that before...

PR #207, opened 2016-05-04
Added lots of checkouts and new versions to 5.1.

mauritsvanrees:
`bin/manage jenkins` goes through the list of packages and reports on changes.
We should do this more often.

Note: I have updated the pin of `z3c.batching` but this not actually used in core Plone.


mauritsvanrees:
Woohoo: we have a Plone Contributors Agreement verifier! Thanks!

mauritsvanrees:
The first test run failed because I updated z3c.form:
https://github.com/zopefoundation/z3c.form/issues/44
I have reverted this part and am now trying again.

jensens:
looks good. do we really depend on collective.js.jqueryui anywhere? sounds scary to have core depend on it.

mauritsvanrees:
No, we are not using collective.js.jqueryui in core. I think we added the pin in the Plone 4 days because various add-ons use it, I think PloneFormGen among them, and there were a lot of people who installed a version that did not work well with their Plone/jquery version. So we added the pin to have a canonical version.
I guess having it in the auto-checkout now has no real effect, since we are indeed not using it. But it serves as a note indicating that this needs a release. Its master branch actually has changes for Plone 5.
I will resolve the conflict of this pull request and then merge.

PR #211, opened 2016-05-06
Bump bootstrap and setuptools

gforcada:


gforcada:
bootstraping works merging.

PR #220, opened 2016-05-27
Check if branch 2.3 of CMFUid works as it is

gforcada:


gforcada:
There seems not to be much difference between 2.2 and 2.3 (they backport from each other) the only real or main difference is some cleanups/refactorings.

Anyone for merging?

PR #221, opened 2016-05-29
Revert "Check if branch 2.3 of CMFUid works as it is"

gforcada:
Reverts plone/buildout.coredev#220

CMFUid tests are not being run at all


gforcada:
Uh... seems I was wrong about that

PR #222, opened 2016-05-29
Add CMFUid to checkouts

gforcada:


gforcada:
Testing and merging https://github.com/plone/buildout.coredev/pull/220 was a bit stupid as CMFUid was not in checkouts.cfg, so no matter which branch sources.cfg is pointing if it actually not checked out. :fearful: 

Let's see if checking out works...

gforcada:
As suspected, branch 2.3 of Products.CMFUid would require some effort on fixing things here and there, but as we might as well want to get rid of it, most probably any effort on that package would be done to remove it rather than fixing it.

PR #224, opened 2016-06-09
update robot versions

thet:


tomgross:
We have massive problems with robot tests and the select2 pattern in Firefox 45. Anyone else facing this issue? Which version of Firefox do we use on jenkins.plone.org?

`AssertionError: WebDriverException: Message: Element is not clickable at point (232.5, 747.7666625976562). Other element would receive the click: <div style="" class="select2-drop-mask" id="select2-drop-mask"></div>
Command duration or timeout: 67 milliseconds
Build info: version: '2.53.0', revision: '35ae25b', time: '2016-03-15 17:00:58'
System info: host: 'localhost.localdomain', ip: '127.0.0.1', os.name: 'Linux', os.arch: 'amd64', os.version: '4.1.13-100.fc21.x86_64', java.version: '1.8.0_65'
Driver info: org.openqa.selenium.firefox.FirefoxDriver
Capabilities [{applicationCacheEnabled=true, rotatable=false, overlappingCheckDisabled=true, handlesAlerts=true, databaseEnabled=true, version=42.0, platform=LINUX, nativeEvents=false, elementScrollBehavior=true, acceptSslCerts=true, webStorageEnabled=true, locationContextEnabled=true, browserName=firefox, takesScreenshot=true, javascriptEnabled=true, cssSelectorsEnabled=true}]
`

gforcada:
@tomgross jenkins.plone.org uses Firefox version 34, we need to update them and with that bump robotframework dependencies as well, but that needs to be carefully tested all together if not things will start to fail all over the place

thet:
I could not get run robot tests locally with firefox at all with the default set of dependencies in buildout.coredev. therefore the motivation to start this pull request. probably only selenium has to be upgaded to support eventually changed web driver API for firefox.

regarding your problem @tomgross - I haven't seen it except that the only one test failure here looks similar. my first idea was that selenium might check if a click like supposed is possible at all and complains about, if some z-index stuff is masking the element? that would be a nice feature actually. i don't know yet, though.

mauritsvanrees:
coredev 4.3 has selenium 2.53.1, which you did in April. I think you tried upgrading 5.0 and 5.1 then as well but ran into similar failures.
Locally on my Mac, I have taken over the selenium 2.53.1 pin in 5.0 and 5.1 (and any other projects), because otherwise Firefox starts up but hangs.

mauritsvanrees:
Can't we set selenium to 2.46.0 in `tests.cfg` and selenium = 2.53.x in `versions.cfg`, like we do in coredev 4.3?

thet:
@mauritsvanrees that sounds like a good idea! Although I also use tests.cfg in some of my own builds, so I would have to add a version fix myself. But that's a special setup.

@mauritsvanrees today I was thinking if we shouldn't just unset the selenium version fix, because browser update more often than Plone releases do. But then we don't have a reproducible build in the future. However, such a reproducible build wouldn't help much anyways, if we would not build it on a reproducible operating system with fixed software versions from the time we made a Plone release...

+1 for your idea.

mauritsvanrees:
I have rebased and updated the pull request and started Jenkins.

thet:
I fixed the robot test failure already in here: https://github.com/plone/Products.CMFPlone/pull/1644

I also removed your change again and try to run again with fixed tests and updated robot versions.
Let's see, if this fixes anything.

thet:
By the way, currently I'm again unable to run robot tests with firefox. It works in chrome. For that, you have to install chromedriver https://sites.google.com/a/chromium.org/chromedriver/downloads and run tests like that:
`ROBOT_BROWSER=chrome ./bin/test --all -m Products.CMFPlone`

PR #227, opened 2016-06-28
Update robot/selenium versions to same as coredev 5.1.

mauritsvanrees:


mauritsvanrees:
http://jenkins.plone.org/job/pull-request-5.0/1252/ has one robot failure, which is expected. That should be fixed with a backport: https://github.com/plone/Products.CMFPlone/pull/1651
I will test them in combination.

PR #230, opened 2016-07-06
Revert plone.app.content

gforcada:
Seems to be breaking some tests.

See http://jenkins.plone.org/job/plone-5.1-python-2.7/569/


gforcada:
Seems that it was due to a broken egg on jenkins nodes..

PR #239, opened 2016-08-25
No longer exclude z3c.form from the tests. [4.3]

mauritsvanrees:


mauritsvanrees:
This gives gives errors like `_p_serial must be an 8-character bytes array`.
This is related to the `persistent` dependency that is pulled in by `z3c.form[test]`.
See also https://community.plone.org/t/zodb3-11-code-creeping-into-plone5/410
On Plone 5 the same dependency gives no problem in the tests, but we still don't want it in production code.
I will try Jenkins once more, because I don't get the error locally when I do `bin/test -s Products.Archetypes`

mauritsvanrees:
screen shot 2016-08-25 at 16 06 27

Nope, apparently fine on the other coredev branches, but a bad idea on Plone 4. Closing.

PR #243, opened 2016-09-26
Add plone.app.event to sources.cfg to test PR with Jenkins

bsuttor:
Goal is testing plone.app.event with Jenkins on Plone 4.3 

http://jenkins.plone.org/job/pull-request-4.3/415/console
https://github.com/plone/plone.app.event/pull/244


jensens:
Huh, this is an edge case I suppose. Ok, lets add it, but I think you need to add both URLs to Jenskins PR job

PR #246, opened 2016-09-29
Checkout z3c.form [5.1]

mauritsvanrees:


jensens:
@mauritsvanrees so this is latest master? Btw., where do we pull in zopefoundation resources?

mauritsvanrees:
No, this is branch 3.2.x of z3c.form. master still has incompatible changes.

The source is listed here: https://github.com/plone/buildout.coredev/blob/5.1/sources.cfg#L148
Or do you mean something else with 'resources'?

PR #253, opened 2016-10-03
just a bunch of updates

jensens:


mauritsvanrees:
Several test failures.
- z3c.form needs to stay on 3.2.9 for now. I think that counts for all or most of the failures. See latest pull requests on z3c.form
- Products.ExternalEditor 2.x is for Zope trunk, so we should stick to 1.x for now.

jensens:
I somehow expected this. I am sure the z3cform problem is something we could solve. 

I think I will expand plone.versioncheck to refer to the tickets or otherwise comments explaining why some package sticks to a specific version. Then we can run it with jenkins every time coredev changes and generate the html output which then could be shown as outcome. Details to be defined. 

What do you think? Also cc @gforcada and others

gforcada:
@jensens if I got you right, you want to get a report from plone.versioncheck on every jenkins run to let us know of newer versions?

I would rather do that once a week or a month basis. There's little to no benefit on getting every single micro update of random dependency X every week. Instead, a monthly update could be much worth.

And for doing that we could extend mr.roboto to create the pull request and start jenkins on its own, so we would only need to accept the pull request.

As for these pins that should not be updated, where will you store that? Inside plone.versioncheck? Then we would need a newer version to remove/add a pin...

jensens:
please lets go to #255 for further discussion (OT here)

PR #254, opened 2016-10-03
update ztk et al

jensens:
experimental, hand-picked every version from change log


jensens:
I close this, too much at once :) lets start smaller. Keep the branch for first.

PR #256, opened 2016-10-04
update zope.component and zope.interface

jensens:
First of a step by step trial how far we may update our ZTK stack in the current setup.

Also experimental usage of plone/plone.versioncheck#25

### DIFF
--- a/versions.cfg
+++ b/versions.cfg
@@ -25,6 +25,8 @@ zdaemon = 4.1.0
 transaction = 1.4.4
 Sphinx = 1.3.6
 zope.app.locales = 3.7.5
+zope.component = 3.12.1
+zope.interface = 4.2.0
 # required for recent z3c.form and chameleon
 zope.pagetemplate = 3.6.3
 # Support for more complex catalog queries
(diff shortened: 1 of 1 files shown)


versions.cfg
```diff
@@ -325,3 +327,11 @@ plone.app.drafts                      = 1.1.1
 plone.app.tiles                       = 2.2.1
 plone.tiles                           = 1.8.0
 plone.app.ldap                        = 1.4.0
+
+[versionannotations]
+# keep this alphabetical please
```
mauritsvanrees, line 329: Keeping it alphabetical seems very good. Can you start by moving zope.component above zope.interface? :-)
Someone can also do that manually before merging, if the pull request is good.

jensens: lol, thanks. ok, need to follow my own rules. this is just experimental.

jensens:
trapped into https://community.plone.org/t/problems-with-mocktestcase-in-plone-dexterity/1484

fixed in https://github.com/plone/plone.mocktestcase/commit/97dbd21bd1116b76f82d843e663d1bd7fd93d6fc

versions.cfg
jensens: done

gforcada:
Isn't that part of the PLIP https://github.com/plone/Products.CMFPlone/issues/1351 ? I mean, is there any little benefit on updating just two out of 200 zope dependencies at a time? Or is this just a testing pull request for your enhancements on plone.versioncheck?

jensens:
It makes it easier to have code we use in pyramid with latest of those packages and plone to work together. Also it does not update to latest, but to latest possible in our current environment.

no, not related to plone.versioncheck

pbauer:
The jenkins job of https://github.com/plone/Products.CMFPlone/issues/1351 (http://jenkins.plone.org/view/PLIPs/job/plip-zope4/) has been failing for a while now. I think it should be more of a priority for us to keep these tests green now that Zope sees more development. Before the conference-sprint I will not have time to work on that.

gforcada:
@pbauer due to that activity (on zope packages) the jenkins job is failing. Basically lots of imports are failing so not even tests can run :-/

It no longer makes sense to target master branches of most packages probably...

PR #258, opened 2016-10-21
Remove Products.i18ntestcase

gforcada:
I don't see it being used anywhere


gforcada:
@tisto any reason to keep that in?

PR #260, opened 2016-10-26
Update sources.cfg, versions.cfg

thet:
Sources.cfg:
- Use Products.ZCatalog 3.0.x branch
- Remove collective.elephantvocabulary
- Remove collective.js.jqueryui
- Use zope.globalrequest 1.2.x branch

versions.cfg:
- Remove collective.elephantvocabulary

The changes with Products.ZCatalog and zope.globalrequest are necessary to allow buildout to run with `auto-checkout = *`

### DIFF
--- a/sources.cfg
+++ b/sources.cfg
@@ -19,13 +19,11 @@ zope_push = git@github.com:zopefoundation
 
 
 [sources]
-AccessControl                       = git ${remotes:zope}/AccessControl.git pushurl=${remotes:zope_push}/AccessControl.git branch=master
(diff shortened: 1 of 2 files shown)


mauritsvanrees:
Actually, collective.elephantvocabulary should not be removed: plone.app.event still uses it. Are you planning to remove it from there?

gforcada:
@mauritsvanrees the dependency on plone.app.event is gone thanks to @davisagli :-) though not released

mauritsvanrees:
But... it is still being pulled in. At the end of a buildout run I get:

```
[versions]
collective.elephantvocabulary = 0.2.5
```

Ah, it is still used in `tests.cfg`. I will remove it from there.

thet:
Thans @mauritsvanrees for checking that!

PR #261, opened 2016-10-26
Get rid of PasswordResetTool

tomgross:


jensens:
you dont need to add a PR to buildout.coredev in order to test multiple packages.

the simple way to do it is to paste the URLs of the involved PRs to the "Build with parameters" field in jenkins, one URL per line:
screenshot from 2016-10-27 14 15 26

PR #262, opened 2016-10-26
Bump qa.cfg versions

gforcada:
I did not update flake8 to 3.0.4 as there are some plugins (at least flake8-print) that still do not work with it

### DIFF
--- a/experimental/qa.cfg
+++ b/experimental/qa.cfg
@@ -30,29 +30,39 @@ flake8-extensions =
     flake8-string-format
 
 [versions]
-flake8 = 2.5.4
-flake8-blind-except = 0.1.0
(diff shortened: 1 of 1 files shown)


gforcada:
@jensens there's no need to run the pull requests jenkins job. The changes in this PR are not related to the main jenkins jobs, but rather to the tab "Pkgs" on jenkins: http://jenkins.plone.org/view/Pkgs/ I still need to double check if it's fine to change to these newer pins or not.

Sorry for not explaining it clearly before.

gforcada:
I've been double checking that all plugins we use do work with the new flake8 3.x versions (specially since the fixes on 3.2.1) and everything seem to work fine, so I'm merging myself.

PR #269, opened 2016-10-30
Downgrade lazr.restfulclient

gforcada:
The latest release is only available on launchpad not in pypi.


gforcada:
Merged manually pinning yet another distribution.

mauritsvanrees:
Been there, done that. I tried the same in March this year: 4ff9c557efad35307e77d538579aecad4aa36fa7

jensens:
annotations for `plone.versioncheck` are working. what about adding a annotation for this dependency :)

PR #271, opened 2016-11-04
Moved p.a.referenceablebehavior and p.f.recurrence to AT tests.

mauritsvanrees:
They both setup sample AT types for testing.
plone.app.referenceablebehavior also needs several AT catalogs, so it cannot work without AT anyway.

When they are in bin/alltests, their tests start failing when we [remove ZopeTestCase from borg.localrole](https://github.com/plone/borg.localrole/pull/5).
I could find no logical reason for that.
There error was that no instance of the sample types could be created.
They ended up in portal_types just fine, but no factory could be found.

This pull request moves both packages to bin/alltests-at.

### DIFF
--- a/tests.cfg
+++ b/tests.cfg
@@ -236,6 +236,8 @@ ATCT =
 AT_plone_app_testing =
     archetypes.multilingual
     plone.app.collection
+    plone.app.referenceablebehavior
+    plone.formwidget.recurrence
     Products.CMFDynamicViewFTI
     Products.contentmigration
 CMF =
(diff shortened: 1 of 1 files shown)


gforcada:
@mauritsvanrees do we want them to be tied up with Archetypes or instead refactor them to be used with dexterity? if the former (Archetypes only) then I would bundle them with the "to remove for Plone 6" and if the later (should support dexterity) then let's convert them to dexterity and keep them on the `plone_app_testing` group

gforcada:
@mauritsvanrees btw thanks for all the work on cleaning up ZopeTestCase tests!!

jensens:
this fixes a problem, we may want to refactor later, I'll merge now.

PR #272, opened 2016-11-07
Products.PasswordResetTool was merged into CMFPlone 5.1

thet:


thet:
Tests are green, got positive review from @jensens - merge myself.

PR #276, opened 2016-11-13
WIPJenkinsfile

gforcada:
This probably does not work at all, it's just a start


tisto:
@gforcada are you aware that we already have a working Jenkinsfile for coredev? https://github.com/plone/jenkins.plone.org/blob/master/flow.groovy

tisto:
Closing. Old Jenkinsfile format...we have a working example already...

PR #282, opened 2016-11-28
Create test runner for Python 3

gforcada:
Only the packages that are known that work on Python 3 should be added.

The idea is to start small,
keep adding packages as they are made compatible with Python 3.

**THERE'S NO NEED TO TEST IT ON THE JENKNS PULL REQUEST JOB**


gforcada:
Merging, the jenkins job is at http://jenkins.plone.org/view/py3/job/py3-test

PR #285, opened 2016-12-19
fix plone.app.contentrules branch to 4.0.x as this PR makes tests not…

thet:
The PR https://github.com/plone/plone.app.contentrules/pull/23 make tests not pass on Plone 5.0, therefore I want to raise the p.a.contentrules version to 4.1 and use 4.0.x branch for Plone 5.0.

/cc @jensens


thet:
no, that was actually not the case. tests did pass!

PR #286, opened 2016-12-19
a bunch of dependency updates

jensens:
lets see what happens on jenkins ... :)

### DIFF
--- a/versions.cfg
+++ b/versions.cfg
@@ -5,7 +5,7 @@ extends = http://dist.plone.org/versions/zopetoolkit-1-0-8-zopeapp-versions.cfg
 [versions]
 ################
 # Zope overrides
-docutils = 0.12
+docutils = 0.13.1
 # Get support for @security decorators
 AccessControl = 3.0.12
 ExtensionClass = 4.1.2
@@ -25,9 +25,12 @@ ZopeUndo = 4.1
 coverage = 4.2
 zdaemon = 4.1.0
 transaction = 1.7.0
-Sphinx = 1.3.6
+transaction = 1.4.4
+Sphinx = 1.5.1
 zope.app.locales = 3.7.5
 zope.component = 3.12.1
+zope.i18n = 4.1.0
+zope.i18nmessageid = 4.0.3
 zope.interface = 4.2.0
 # required for recent z3c.form and chameleon
 zope.pagetemplate = 3.6.3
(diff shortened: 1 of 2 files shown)


jensens:
the zope.i18nmessageid test problem is addressed here plone/plone.supermodel#18 and here plone/plone.schemaeditor#45

versions.cfg
```diff
@@ -40,14 +43,15 @@ Products.Sessions = 3.0
 
 #############
 # Build Tools
-
+
```
gforcada: @jensens would you mind removing this?

versions.cfg
```diff
@@ -25,9 +25,12 @@ ZopeUndo = 4.1
 coverage = 4.2
 zdaemon = 4.1.0
 transaction = 1.4.4
-Sphinx = 1.3.6
+Sphinx = 1.5.1
+unittest2 = 1.1.0
```
gforcada: @jensens why is unittest2 reintroduced? I've been trying to remove it for a while

versions.cfg
```diff
@@ -40,14 +43,15 @@ Products.Sessions = 3.0
...
-
+ 
 # Basics
 # !! keep in sync with requirements.txt !!
 setuptools = 26.1.1
 zc.buildout = 2.5.3
 
 # recipes and extensions
 collective.recipe.omelette = 0.16
+collective.recipe.sphinxbuilder = 1.0
```
gforcada, line 54: @jensens do you remember who needs it?

gforcada (commented):
Nice work, I added some questions though :-)

versions.cfg
jensens: It is used in the main ``buildout.cfg`` of ``buildout.coredev``

versions.cfg
jensens: it is used by robotsuite and http://dist.plone.org/versions/zopetoolkit-1-0-8-ztk-versions.cfg pins a very old 0.5.1

versions.cfg
jensens: sure

jensens:
- [x] merge plone/Products.Archetypes#72
- [x] wait for release of ``robotsuite`` and get rid of unittest2 (and its dependencies)

versions.cfg
jensens: see also collective/robotsuite#12 - then we can unpin. wait for release

mauritsvanrees:
For the record, the `html5lib` version increase gave errors when using `zest.releaser`. I fixed that in ec2cb40f6905ab049ce38dd896f18693e91b14a0. I did update the version as much as possible, effectively from 1.0b7 to 1.0b8.

jensens:
html5lib and its versioning crap is a big PITA. Sorry for the inconvenience.

PR #291, opened 2017-01-12
BTrees 4.4.0 with None support again

jensens:
Let's try BTrees 4.4.0 with None support.

From its changelog:

>     Allow None as a special key (sorted smaller than all others).
> 
>     This is a bit of a return to BTrees 3 behavior in that Nones are allowed as keys again. Other objects with default ordering are still not allowed as keys.
>

### DIFF
--- a/versions.cfg
+++ b/versions.cfg
@@ -21,6 +21,7 @@ MultiMapping = 3.0
 ZConfig = 3.1.0
 ZODB = 3.10.7
 ZopeUndo = 4.1
+BTrees = 4.4.0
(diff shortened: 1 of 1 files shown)


davisagli:
This change doesn't make sense since we are still using ZODB3, which bundles BTrees. We only need/want a separate BTrees pin when we upgrade to ZODB 4. (I tested that locally and it seems to work.)

Btw I notice we have a pin for ZODB = 3.10.7 here, which does not actually exist on PyPI.

PR #298, opened 2017-01-25
use newer Chameleon 3.0

jensens:
let's wait for the tests


jensens:
five.pt does not like it

jensens:
see also plone/Products.CMFPlone#1804 and plone/Products.CMFPlone#1351

jensens:
This does not work in naked Plone 5.1 (may work with some lovel) but works with ``plip-zope4.cfg`` (where it is now pinned). So I close here.

PR #302, opened 2017-02-01
Update AccessControl to 3.0.12, released December 2015

mauritsvanrees:
This then requires an update of ExtensionClass and Acquisition to much newer versions.
So this may be tricky.  We had some problems getting this on Plone 5.
See what Jenkins says.

### DIFF
--- a/checkouts.cfg
+++ b/checkouts.cfg
@@ -11,3 +11,4 @@ auto-checkout =
     Products.statusmessages
     Products.PlacelessTranslationService
     Zope2
+    five.intid
--- a/sources.cfg
+++ b/sources.cfg
@@ -20,6 +20,7 @@ zope_push = git@github.com:zopefoundation
 [sources]
 Zope2                               = git ${remotes:zope}/Zope.git pushurl=${remotes:zope_push}/Zope.git branch=2.13
 ZODB3                               = git ${remotes:zope}/ZODB.git pushurl=${remotes:zope_push}/ZODB.git branch=3.10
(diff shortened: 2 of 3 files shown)


gforcada:
Any reason to try to merge that on 4.3 rather than 5.1?

mauritsvanrees:
Plone 5 already had this for a long while.

gforcada:
Oh I see, sorry for the noise

mauritsvanrees:
It is basically #158 reworked for 4.3. But it may be too many version increases for 4.3.
The first Jenkins run failed with just a few errors. I force pushed some more changes, so let's try again. But my hopes are low.

mauritsvanrees:
The `five.intid` error is fixed, but `plone.dexterity` still fails with the same error. So the increased version did not help.

It fails in `plone.dexterity.tests.test_webdav.TestFolderDataResource.test_PROPFIND`, where in the result we are missing a few lines:

```
4a5,10
> <d:propstat xmlns:n="http://www.zope.org/propsets/default">
>   <d:prop>
>   <n:title>Container</n:title>
>   </d:prop>
>   <d:status>HTTP/1.1 200 OK</d:status>
> </d:propstat>
21a28
>   <n:getlastmodified>...</n:getlastmodified>
```

And in `plone.dexterity.tests.test_webdav.TestFolderDataResource.test_PROPPATCH` we get `Method not supported for this resource.`

I'm not really sure what is going on there.

gforcada:
@mauritsvanrees do you still want to get that merged? Please re-open if so.

PR #307, opened 2017-03-04
Mr.developer 1.37

fschulze:
Testing 1.37b1 first. It should provide more info on git operations.


fschulze:
It seems I have to push a branch into the original repo, a fork doesn't seem to work.

PR #309, opened 2017-03-09
4.3: use plone.app.linkintegrity 1.5.10

mauritsvanrees:
This reverts commit 5e708c61c76603c939951ee96a0f4f4034d3447e.

Locally, 1.5.8 works. But 1.5.9 failed, so I rereleased this as 1.5.10, but that fails as well:

```
Getting distribution for 'plone.app.linkintegrity==1.5.10'.
While:
  Updating instance.
  Getting distribution for 'plone.app.linkintegrity==1.5.10'.
Error: Couldn't find a distribution for 'plone.app.linkintegrity==1.5.10'.
```

I used Python 2.6. But then I tried 2.7 but that failed in exactly the same way. `bin/buildout -vv` does not give any extra info. This is on a Mac.
In a virtualenv I can do `pip install --no-deps plone.app.linkintegrity==1.5.10` just fine.

I have no idea what is going on here. But let's see if Jenkins has the same problem.

If the build is green and you like to merge this: please try it yourself on your local machine first, especially if you have a Mac.


mauritsvanrees:
Well, Jenkins tests pass, which I guess is good.

Locally on my Mac the following simple buildout fails:

```
[buildout]
extends = http://dist.plone.org/release/4.3.11/versions.cfg
parts = test

[versions]
plone.app.linkintegrity = 1.5.10

[test]
recipe = zc.recipe.egg
eggs = plone.app.linkintegrity
```

```
$ bin/buildout 
Installing test.
Getting distribution for 'plone.app.linkintegrity==1.5.10'.
While:
  Installing test.
  Getting distribution for 'plone.app.linkintegrity==1.5.10'.
Error: Couldn't find a distribution for 'plone.app.linkintegrity==1.5.10'.
```

mauritsvanrees:
4.3 uses an old `zc.buildout` and `setuptools`, so I tried with the most recent ones, and some versions in between, but that did not help.

```
zc.buildout = 2.9.2
setuptools = 34.3.1
```

mauritsvanrees:
I tried it on a Linux machine, and there it failed to. But Jenkins passes, and in the console I see that it does get the new linkintegrity version.

I am puzzled...

gforcada:
I can only say that it works for me™️ locally :-/ I'm on fedora linux for that matter.

mauritsvanrees:
And... now it works for me locally too, both 1.5.9 and 1.5.10.
I don't know why, and I probably don't want to know...
I'll merge.

PR #311, opened 2017-03-21
z3c.form 3.4.0 (2nd trial)

jensens:
Recreates (Reverts the revert of) plone/buildout.coredev#310 

Same as plone/buildout.coredev#278, see details there.


jensens:
There have been failures in ``plone.autoform`` (at least randomly). We need to investigate here.

Issue #235, opened 2016-08-19
Update coverage


idgserpro:
Today is used coverage 3.5.2. However, the newer createcoverage requires [coverage>=3.7](https://github.com/reinout/createcoverage/blob/1.5/setup.py#L13). When used createcoverage we get the error:

``` bash
Installing createcoverage.
The constraint, 3.5.2, is not consistent with the requirement, 'coverage>=3.7'.
While:
  Installing createcoverage.
Error: Bad constraint 3.5.2 coverage>=3.7
```

Who pin coverage is [zopetoolkit-1-0-8](https://github.com/zopefoundation/zopetoolkit/blob/1.0.8/ztk-versions.cfg#L105), that is used [here](https://github.com/plone/buildout.coredev/blob/a8c49fb2c794a8ccc4ed1dd7889729de722b5697/versions.cfg#L2). It would not be appropriate to update the zopetoolkit?

idgserpro:
@esteele?

mauritsvanrees:
Seems a good idea to me. I have opened a pull request: https://github.com/plone/buildout.coredev/pull/236

That is for 5.1. Probably good for 5.0 too. For 4.3 I am slightly more hesitant, but.. well, why not.

@hvelarde I think you fixed quite some Travis setup in packages to use the right coverage/createcoverage combination. A newer coverage pin will be easier for new packages. But I wonder if existing packages would then need yet more Travis changes. Do you maybe have a preference for sticking to the old version in coredev 4.3?

I can never remember how I usually fix this in packages: pin an older createcoverage, or pin a newer coverage. I guess I usually pin a newer coverage.

mauritsvanrees:
For now, I have merged to 5.1 and have applied the same change to 5.0.

idgserpro:
@mauritsvanrees I'm +1 to update in Plone 4.3, too. [coveralls](https://github.com/plone/bobtemplates.plone/blob/master/bobtemplates/plone_addon/.travis.yml.bob#L18) only works with coverage>=4.0. See: https://github.com/mozilla/sugardough/issues/94

hvelarde:
@mauritsvanrees I did it manually removing the pinning on coverage; there's no problem on updating on 4.3, this is what I normally do:

```
[versions]
# use latest version of coverage
coverage =
```

not the best thing to do but...

mauritsvanrees:
I have added the pin to 4.3 as well.

[...]

PR #312, opened 2017-03-21
Revert to coverage 3.x

MatthewWilkes:
Undo the change to coverage in e779e2b21a891a69fb56cf7cb09b6d1ddd4823ae as this creates an impossible dependency loop with z3c.coverage 2.0.3, which requires coverage < 4.

### DIFF
--- a/versions.cfg
+++ b/versions.cfg
@@ -14,7 +14,7 @@ Acquisition = 4.2.2
 DateTime = 4.0.1
 Products.BTreeFolder2 = 2.14.0
 # Override until ztk is updated
-coverage = 4.2
+coverage = 3.7.1
 Sphinx = 1.3.6
 # required for recent z3c.form and chameleon
 zope.pagetemplate = 3.6.3


MatthewWilkes:
@mauritsvanrees FYI, as this was your change originally. Not sure if you have extra information that informs this.

mauritsvanrees:
See https://github.com/plone/buildout.coredev/issues/235 for a discussion on versions of coverage related packages. Summary from quickly glancing at it again:
- that issue was initially about not using the even older `coverage 3.5.1`
- `coveralls` needs `coverage 4` or higher
- `z3c.form` was requiring `z3c.coverage` in its `test` dependencies, which is why we needed to be careful about versions.
- instead of `z3c.coverage` you can use `createcoverage`, which I believe is what more community packages are using these days, for example via https://github.com/collective/buildout.plonetest/blob/master/qa.cfg. Also, it is created/maintained by my brother. :-)

MatthewWilkes:
Well, currently it's not possible to build Plone if you're using z3c.coverage in the buildout, so if we need coverage4 it seems all we can do is pin z3c.coverage to be an old version that erroneously doesn't check coverage versions?

mauritsvanrees:
Yeah, `z3c.coverage 2.0.2` would be the latest that 'works'.

icemac:
The `z3c.coverage` of `z3c.form` was removed in 3.2.11 resp. 3.3.0. So this package should not cause a problem.

jensens:
I do not think the downgrade is what we want, we want to get rid of the usage of z3c.coverage.
So my opinion: :-1:

gforcada:
It's been almost a year... I'm closing it, but if someone feels that it should still keep open, do so 🙂

PR #318, opened 2017-03-31
Use current ZODB (HT @jimfulton)

hvelarde:


hvelarde:
@jimfulton tests are failing here and seem there are some incompatibilities; I see the following issues repeated many times:

* `TypeError: unhashable type: 'Missing.Value'`
* `AttributeError: 'module' object has no attribute 'TestCase'`

I have no idea if this will be difficult to fix.

CC @plone/framework-team

jensens:
I don't think this should go into 4.3 - but this is more release management related than everything else.  
cc @esteele

gforcada:
@hvelarde we can not backport every single new feature back to Plone 4.3, if it's supposed to be a LTS no major changes should be added here, and albeit is *only* about bumping some dependencies, those are quite important, so I'm -1 on merging it on 4.3

That does not mean that it is *impossible* to have ZODB5 in Plone 4.3, just that from the release management point of view I would not encourage, that when updating from 4.3.13 to 4.3.14 you have fixes here, fixes there, and oh, btw. not a major, but a two major releases of ZODB bump (3.x to 5.x).

PR #319, opened 2017-03-31
Use current ZODB (HT @jimfulton)

hvelarde:


jensens:
Same as with 4.3: I don't think this should go into 5.0 - but this is more release management related than everything else.
IMO 5.1 is the right place for the ZODB version update.
cc @esteele

jensens:
this has too many implications, as we saw on 5.1

PR #317, opened 2017-03-31
use current ZODB

jimfulton:
bin/alltests passes for me.

I'm not set up to run the selenium tests, but am told that they're be tun automatically on a PR.

PR #320, opened 2017-04-05
Use current ZODB (2nd trial)

jensens:
Tests were red after merge of #317 - we need to figure out what happened - they were green on PR before.

### DIFF
--- a/versions.cfg
+++ b/versions.cfg
@@ -21,17 +21,22 @@ MultiMapping = 3.0
 ZConfig = 3.1.0
 zLOG = 3.0
 ZopeUndo = 4.1
-BTrees = 4.4.1
 # Overrides until ztk is updated
 coverage = 4.3.4
 zdaemon = 4.2.0
-transaction = 1.7.0
 Sphinx = 1.5.2
 zope.app.locales = 3.7.5
 zope.component = 3.12.1
 zope.i18n = 4.1.0
 zope.i18nmessageid = 4.0.3
 zope.interface = 4.2.0
+# ZODB and related
+ZODB = 5.2.0
+ZEO = 5.1.0
+ZODB3 = 3.11.0
+tempstorage = 4.0
+BTrees = 4.4.1
+transaction = 2.1.2
 # required for recent z3c.form and chameleon
 zope.pagetemplate = 3.6.3
 # Support for more complex catalog queries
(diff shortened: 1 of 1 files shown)


gforcada:
Do you have the link to the jenkins job that failed?

jensens:
Huh, after looking at it, it seems the merge before was the cause of the red Jenkins: http://jenkins.plone.org/job/plone-5.1-python-2.7/1753/
But there were no changes and the next one, finally green, was the ZODB merge:  http://jenkins.plone.org/job/plone-5.1-python-2.7/1754/
Overall, this is a bit confusing.

jensens:
Now, the PR fails. I really have no idea what's going on here.

In parallel, I have the tests running local (w/o robot tests) and they seem to pass (still running, but some of the failed ones passed).

jensens:
I start the tests again with more version pins (ZODB dependencies were not pinned). 

Unrelated, I also added ZRS and RelStorage to the ecosystem pins in order to help future users to not use outdated versions.

jensens:
btw., local on my machine all tests are passing (w/o having robot tests run).

jensens:
It fails again, this looks like a machine/setup/jenkins specific problem. I am a bit lost here.

gforcada:
I have no idea, but someting looks really broken: http://jenkins.plone.org/job/pull-request-5.1/1456/consoleFull

gforcada:
Given that the main jobs are green, it has to be due to the changes here...

jensens:
@gforcada given that http://jenkins.plone.org/job/plone-5.1-python-2.7/1754/ was green: I am confused. May you explain this?

jensens:
@gforcada some time (about a year IIRC) we had a similar situation where one node was broken and produced total failure while the others were working. No idea if this is related. 

Also, the broken build one commit before the ZODB5 merge in 5.1 branch and the subsequent green build of the merge signaling there is something wrong with Jenkins.

gforcada:
Yes, that's true, node1 was the problematic one, but this last error happened on node5, which together with node6, are the latest nodes we have, and they should be fine...

jensens:
@gforcada I am still curious why  http://jenkins.plone.org/job/plone-5.1-python-2.7/1754/ was green (it was this ZODB change). Any chance to figure out? There must be something unstable in the stack.

PR #321, opened 2017-04-07
use newest plone.versioncheck / dev version on buildout.coredev

loechel:


versions.cfg
```diff
@@ -61,7 +61,7 @@ plone.recipe.alltests = 1.5
 plone.recipe.zeoserver = 1.3
 plone.recipe.zope2instance = 4.2.22
 plone.releaser = 1.5.4
-plone.versioncheck = 1.5.1
+plone.versioncheck =
```
gforcada: Is that meant to try the dev version of plone.versioncheck only or meant to be merged? I would strongly oppose to leave a version unpinned on any Plone official release. If the former (just to test) it would be nice if you prefix this pull request with ``[WIP]`` on its title.

jensens: Indeed, please keep the version pinned. If plone.versioncheck is added to auto-checkout it overrides the version pin anyway.

Also I prefer to not have WIP in title but a proper use of the labels ``status: in-progress``

jensens (dismissed):
Please do not un-pin the version.

versions.cfg
loechel: resolved

PR #329, opened 2017-04-29
Revert setuptools to 26.1.1

gforcada:
Anything prior to setuptools 33.1.1 should be fine.

Plone uses basic features of setuptools and as our Jenkins CI environment
is already pining 26.1.1 is easier to stick to this version.

As soon as PLIP 2001 https://github.com/plone/Products.CMFPlone/issues/2001 is approved and merged,
we can happily move to virtualenv which solves the problem completely.

@plone/release-team @plone/framework-team opinions? I literally spent +6 hours trying to get a new node on jenkins and failing because of this newer setuptools.

Is not failing on the current nodes because a dependency (``appdirs``) of setuptools is already installed and thus does not have to fetch it (that's the actual problem).

If this is merged, I would like to ask everyone (specially @jensens) to **NOT** bump setuptools versions without first checking with the testing team. It gives a nice feeling to use the latest versions of packages, but if that has to cause massive problems on other ends at random times, then, please don't do it.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,6 +1,2 @@
-appdirs==1.4.2
-packaging==16.8
-pyparsing==2.1.10
-setuptools==34.3.0
-six==1.10.0
+setuptools==33.1.1
 zc.buildout==2.8.0
--- a/versions.cfg
+++ b/versions.cfg
@@ -51,11 +51,7 @@ trollius = 2.1
 
 # Basics
 # !! keep in sync with requirements.txt !!
-appdirs = 1.4.2
-packaging = 16.8
-pyparsing = 2.1.10
-setuptools = 34.3.0
-six = 1.10.0
+setuptools = 33.1.1
 zc.buildout = 2.8.0
 
 # recipes and extensions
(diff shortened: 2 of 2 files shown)


mauritsvanrees:
Thanks for trying to get a new Jenkins node working!
Since it apparently causes so much trouble on the testing team, I am fine with this change.

I would be inclined to pin `setuptools` to `33.1.1` as that is the last version without the extra dependencies, and lots of [fixes](https://github.com/pypa/setuptools/blob/v33.1.1/CHANGES.rst) have been done since `26.1.1`, though nothing specifically jumps out for me. Would updating the Jenkins scripts to a `33.1.1` be a simple search and replace, or is deploying such a change a headache? (BTW, I see some scripts use 21.0.0, probably better to keep all versions the same.)

For clarity I would rather remove the other dependencies from `requirements.txt` and only keep the pins of `setuptools` and `zc.buildout`.

BTW, I see no need to call out individuals here.

jensens:
I agree with @mauritsvanrees

PR #333, opened 2017-05-03
Revert Products.CMFCore to 2.2.10. [5.0]

mauritsvanrees:
We do not want the merged collective.indexing in 5.0.
See https://github.com/plone/Products.CMFPlone/issues/2042


gforcada:
@mauritsvanrees thanks for noticing that! I was wondering what version pin should that be, but I couldn't find any right away (I have some problems at work as well :-/)

mauritsvanrees:
That is weird: the [console log of Jenkins](http://jenkins.plone.org/job/pull-request-5.0/19/console) shows it is still using CMFCore 2.2.11 and it has lots and lots of failures.
Let me try again.

mauritsvanrees:
Still the same problem: the pull request pins Products.CMFCore back to 2.2.10, but the console shows it is still using 2.2.11.

@plone/testing-team: In the console I do not see anything that looks like a switch to the coredev pull request branch. Did that get lost in recent Jenkins changes?

mauritsvanrees:
The [`pr-tests.sh` script](https://github.com/plone/jenkins.plone.org/blob/master/scripts/pr-tests.sh#L7-L10) checks if the `COREDEV` variable is 1, and then does `git checkout $BRANCH`.

In the [Jenkins job variables](http://jenkins.plone.org/job/pull-request-5.0/20/injectedEnvVars/) I see `COREDEV=1` so that seems good. I see `PKGS` with value `BRANCH=cmfcore-revert-50` which looks strange, but maybe that is how it always is.

mauritsvanrees:
Ah, I think this Jenkins [commit from April](https://github.com/plone/jenkins.plone.org/commit/dbf478a17a703eec8f07d163c8612903887a0849) is the cause. It writes `PKGS = ` and forgets to add a newline.

gforcada:
I merged that pull request on jenkins.plone.org (thanks @mauritsvanrees !) and updated the jobs. A new pull request job is running :-)

gforcada:
Green! Thanks again @mauritsvanrees ! 🙇

mauritsvanrees:
Thank you @gforcada for keeping Jenkins up and running!

PR #340, opened 2017-05-15
update z3c.form to latest 4.x

jensens:
3rd trial


gforcada:
@jensens This can be closed, right? As per #384

jensens:
Changed title and reopened.

gforcada:
on 5.1 or 5.2?

jensens:
Good point, I think we should go 5.2 here

PR #342, opened 2017-05-30
new plone.recipe.zeoserver needs new zope.mkzeoinstance. + ZODB minor…

thet:
… upd


thet:
Fixes zeoserver start script.

PR #343, opened 2017-06-15
Use Zope4 version pins for Py3 buildout

tomgross:


gforcada:
LGTM, running the tests locally I get an error, did you get all tests passing?

```
File "/coredev-py3/src/plone.event/plone/event/utils.py", line 430, in plone.event.utils.pydt
Failed example:
    pydt(DateTime('2005/11/07 18:00:00 Brazil/East'))
Expected:
    datetime.datetime(2005, 11, 7, 18, 0, tzinfo=<DstTzInfo 'Brazil/East' BRST-1 day, 22:00:00 DST>)
Got:
    datetime.datetime(2005, 11, 7, 18, 0, tzinfo=<DstTzInfo 'Brazil/East' -02-1 day, 22:00:00 DST>)
```

tomgross:
I have the same failure with and without the change on my Fedora. I thought it was my machine.

jensens:
Looks like the tests check needs to be improved. Annoying, the tzinfo representation should be the same, but that is out of the scope of our tests.

tomgross:
I guess this is obsolete.

PR #350, opened 2017-06-28
Update code QA package versions

thet:
/cc @gforcada 

seems an updated version of flake8-isort among others need those updated versions.
not an issue with buildout.coredev 5.1, as these version pins aren't present there.


gforcada:
@thet I'm curious about who is using these version pins actually. At jenkins.plone.org we do run some QA checks with flake8,but rather than tests.cfg they use experimental/qa.cfg...

gforcada (approved):
Looks sane, but as noted on the other comment, I'm not sure who is using these version pins, so is hard to say what can break

thet:
@gforcada tnx for the review, I'll merge.

In 5.1 these version pins are removed.
Anyway, I'd prefer to have them in versions.cfg - I even did not know that there are version pins in test.cfg...

PR #353, opened 2017-07-12
Remove plone.openid and plone.app.openid

gforcada:
PLIP 1659 was approved and merged some time ago already, see:
https://github.com/plone/Products.CMFPlone/issues/1659

Still, while a regular instance of Plone 5.1 does not have openid related packages,
here on buildout.coredev and specially on tests there were these references now removed.


gforcada:
*as expected* tests do pass 🎉  anyone willing to merge?

PR #363, opened 2017-09-13
we need to depend on the 1.11.x branch for Plone 5.1

thet:


thet:
OK, i thought we need that branch-pin, but we do not. Closing this PR.

PR #369, opened 2017-09-20
update Chameleon to 3.1

jensens:
I do not expect this to work, but we should not forget about it.


jensens:
I don't think it's worth to upgrade this for Plone 5.1

PR #372, opened 2017-09-20
update zope: .component, .interface, .schema and .proxy

jensens:


versions.cfg
```diff
@@ -33,19 +33,21 @@ six = 1.10.0
 tempstorage = 4.0
 zc.lockfile = 1.2.1
 zope.app.locales = 3.7.5
-zope.component = 4.3.0
-zope.contenttype = 4.2.0
 zope.browser = 2.1.0
 zope.browsermenu = 4.2
 zope.browserpage = 4.1.0
 zope.browserresource = 4.1.0
+zope.component = 4.4.0
```
tomgross, line 40: Don't we stay on 3.12.1 until https://github.com/zopefoundation/zope.component/pull/33 is released?

jensens: Hmm, we have been on 4.3.0 already

PR #376, opened 2017-10-18
update ZopeUndo in order to fix broken undo form

jensens:


jensens:
related to plone/Products.CMFPlone#2175

PR #377, opened 2017-10-21
move wsgi parts to core.cfg

tschorr:
Move the WSGI parts added with PLIP https://github.com/plone/Products.CMFPlone/issues/1351 into core.cfg.


tschorr:
already done in https://github.com/plone/buildout.coredev/commit/f100f861e9c19a62937497e10e40d6bb58cab1ca

PR #384, opened 2017-12-20
use latest 3.x of z3c.form after semver problem was solved.

jensens:


jensens:
same to apply to 5.2 branch

mauritsvanrees (approved):
LGTM. There are no code changes between 3.2.11 and 3.6 when I compare the tags in git. But I just started a Jenkins job to be sure. Let's wait for that to finish.

PR #395, opened 2018-01-27
minor upgrades

jensens:


mauritsvanrees:
It fails because of the GenericSetup increase. It needs a merge of https://github.com/zopefoundation/Products.GenericSetup/pull/52
But the other version increases seem fine.

PR #399, opened 2018-01-30
remove dependency on PlacelessTranslationService

jensens:
merge with 
- plone.i18n
- plone.app.testing

### DIFF
--- a/tests.cfg
+++ b/tests.cfg
@@ -116,7 +116,6 @@ test-eggs =
     Products.GenericSetup
     Products.Marshall
     Products.MimetypesRegistry
-    Products.PlacelessTranslationService
     Products.PlonePAS
     Products.PluggableAuthService
     Products.PluginRegistry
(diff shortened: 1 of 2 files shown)


gforcada:
@jensens no idea right now, I will try to debug it

gforcada:
@jensens now I see: https://github.com/plone/jenkins.plone.org/blob/master/jobs/scripts/pr-get-info.py#L145

switching branches and at the same time modifying ``sources.cfg`` does not work really well.

One option is the changes on sources.cfg on this branch itself so only this pull request needs to be tested, but rather than merging it after it succeeds, only the part that is not about sources.cfg could be merged (and of course the other branches).

Fixing the scripts to do it afterwards is not a 5 minutes job and I'm a bit busy right now 😞

jensens:
@gforcada, Thanks for analyzing the problem! I can remove the line from ``sources.cfg`` afterwards, it is not a problem. I try now.

PR #406, opened 2018-02-03
Update test isolation docs

gforcada:


gforcada:
Merged manually but only the comments.

Issue #393, opened 2018-01-27
deduplicate Contributor's Agreement info


tkimnguyen:
re: https://github.com/plone/documentation/issues/672 and https://github.com/plone/buildout.coredev/pull/386 

I think the link https://plone.org/foundation/contributors-agreement/agreement.pdf on the page https://docs.plone.org/develop/coredev/docs/contributors_agreement_explained.html should point instead to https://plone.org/foundation/contributors-agreement/ rather than directly to the PDF, and the information here on docs.plone.org about the contributor agreement should not be duplicated here; it is best left on plone.org

mrsaicharan1:
I would like to fix this.

mrsaicharan1:
Could you provide me a link to the repo where I could fix the broken link?

mrsaicharan1:
I've fixed the broken link issue. Could you review this on https://github.com/plone/buildout.coredev/pull/411?
And also, I think it would be better to provide a download link to the contributors' agreement pdf on the page which I've provided the link to.

gforcada:
@mrsaicharan1 the link to the PDF is already there on that page :-)

fredvd:
Fixed, referenced pr's are merged afaics.

PR #411, opened 2018-02-10
Fix broken link

mrsaicharan1:
Fix for the link
to the agreement pdf.
Fixes https://github.com/plone/buildout.coredev/issues/393

### DIFF
--- a/docs/contributors_agreement_explained.rst
+++ b/docs/contributors_agreement_explained.rst
@@ -7,7 +7,7 @@ which assigns copyright in the code to the Plone Foundation, the non-profit orga
 
 This document explains the purposes of this, along with questions and answers about what this means.
 
-You can download *The Plone Contributor Agreement* from `plone.org <https://plone.org/foundation/contributors-agreement/agreement.pdf>`_.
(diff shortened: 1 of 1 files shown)


gforcada:
@mrsaicharan1 thanks for providing the patch!

Unfortunately, if you remove the angle brackets `<` `>` the whole link is pointing to nowhere. Removing the PDF file name so that it points to the page instead is ok, but keep the angle brackets please.

If you fix that we could merge it already, but that would not fulfill #393 as there it is asked to remove the duplicated parts that are already on the plone.org page.

One last minor thing: as soon as we merge this one, would you mind to repeat the same pull request for the branch 5.2?

mrsaicharan1:
Yes, you're right! My bad! I will do the necessary changes and get back to you. 
And yeah, I will do the same for branch 5.2.

mrsaicharan1:
I have moved this to another new PR https://github.com/plone/buildout.coredev/pull/412. Also, should I remove the duplicated parts from docs.plone.org page or plone.org page? Could you provide me the exact link?

PR #414, opened 2018-02-13
Fix broken link

mrsaicharan1:
Fix link to the 
agreement page.


mrsaicharan1:
Yes, I should've just added the changes in the previous PR. Thanks for pointing that out @gforcada ! 
I will keep that in mind while contributing in the other repos as well!

PR #415, opened 2018-02-20
Update robotframework-selenium2library

ale-rt:
robotframework-selenium2library = 3.0.0 is compatible with Python 3

PR #407, opened 2018-02-08
 Bump robotframework and its dependencies [5.2]

gforcada:
Firefox 50+ is supported now. Copied from https://github.com/plone/plone.app.robotframework/pull/71

PR #416, opened 2018-02-20
Update lazr.restfulclient

ale-rt:
This package should be Python 3 compatible since version 0.13.4 (2014-12-05), see https://pypi.org/project/lazr.restfulclient/#description


gforcada:
See this comment https://github.com/plone/plone.releaser/blob/97082e05eab3425966bbf845d5616860fe65d2cf/setup.py#L47-L49 we might not want to update it right now.

This package is only needed when using plone.releaser, i.e. for the release team to cut releases and update buildout.coredev and so on.

Looking at that issue pointed on the comment, it might work, a new release was pushed to pypi and it should fix the problem.

Le'ts try it :-)

ale-rt:
I needed it to have the 5.2 branch run on Python 3.
Also this is needed #415, but it might be replaced by #407 (which is currently failing).

Anyway thanks for merging :)

PR #417, opened 2018-03-01
remove unittest2

kakshay21:
see https://github.com/plone/Products.CMFPlone/issues/1882

### DIFF
--- a/tests.cfg
+++ b/tests.cfg
@@ -187,7 +187,6 @@ smmap2 = 2.0.3
 testresources = 2.0.1
 traceback2 = 1.4.0
 twine = 1.8.1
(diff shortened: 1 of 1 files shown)


gforcada:
Would you mind removing traceback2 and linecache2? And if it's not too much to ask, could you rebase this branch on top of 5.2 instead? 5.1 is about to get released (expected today actually), so cleanup tasks are better to be moved on the next release. Thanks in advance!

kakshay21:
@gforcada what about traceback2? Should I remove that as well?

kakshay21:
there are too many merge conflicts, opening a new PR looks a viable option. @gforcada

PR #432, opened 2018-03-20
include plone.dexterity translations

erral:


erral:
I messed up with the translations, so I retired the former PRs and issued new ones.

PR #436, opened 2018-03-20
Workaround for https://github.com/pypa/pip/issues/5081#issuecomment-374563426

thet:


mauritsvanrees:
Might be okay, but this would also mean that people who use this file and were using say `pip 7.1.2` will now suddenly be upgraded to 9.0.1. And if you installed 9.0.2 because for some Mac versions this apparently solves a problem, then you would get the old version again.

What does this solve? I have read the pip issue you point to, and the traceback you posted there, but how can I reproduce this? I created a virtualenv in coredev 5.1 and installed the apparently broken `pip 9.0.2` in there, and ran buildout, and the instance started up just fine. `pip` is not included in the `bin/instance` script.

Do you have an add-on that has `pip` in its dependencies? Couldn't that add-on be fixed?

I wonder if something could be changed in `z3c.autoinclude` to fix this, but I guess it would be tricky.

thet:
I could only reproduce this in a docker container setup. I don't really know what causes the pip import. Actually it seems that the problem raises when urllib3 is imported.
I'd have to investigate further - for now my own solution is to pin to 9.0.1 myself.

jensens:
now, with pip 9.0.3 - this is no longer needed.

PR #437, opened 2018-03-20
Workaround for https://github.com/pypa/pip/issues/5081#issuecomment-374563426

thet:


jensens:
now, with pip 9.0.3 - this is no longer needed.

PR #440, opened 2018-03-20
include new packages in plone domain

erral:
Include new packages in plone domain also in 5.2 branch:

plone.app.caching: https://github.com/plone/buildout.coredev/pull/439
plone.dexterity: https://github.com/plone/buildout.coredev/pull/435
plone.app.multilingual: https://github.com/plone/buildout.coredev/pull/429


vincentfretin:
Great! Indeed, it should be in sync with 5.1 branch.

PR #444, opened 2018-03-26
update versions

jensens:


versions.cfg
```diff
@@ -151,7 +151,7 @@ five.customerize                      = 1.1
 five.intid                            = 1.1.2
 icalendar                             = 4.0.1
 mockup                                = 2.7.0
-Plone                                 = 5.1rc2
+Plone                                 = 5.1.1
```
gforcada, line 154: No idea about the other bumps, but please, leave this one for the release team.

jensens: It has no effect on coredev 5.2 and just gets in sync with the 5.1 branch. So no need to worry. And in past and now the release team never cared about dependency versions. So this all is fine and as usual. Just tests need to go green with the updates.

jensens:
Note to self, need to check dateutil and pytz updates. One or both are breaking tests.

jensens:
Excluded both, pytz and date-utils from this PR. New PR for those are in preparation.

PR #449, opened 2018-04-03
remove p.a.ldap 

jensens:
no longer officially supported


ale-rt:
Should this be moved in the ecosystem.cfg?

jensens:
Looks like we removed it from 5.2 anyway. I think we keep 5.1 as its in maintenance now.

PR #451, opened 2018-04-04
More Py3 merges

tomgross:


pbauer:
😆 I just did exactly the same (except fot the auto-checkouts thing)

Issue #462, opened 2018-04-25
Add version file for Zope2 2.13.28/29 on dist.plone.org


mauritsvanrees:
For coredev 4.3, 5.0 and 5.1 we need a new versions file at http://dist.plone.org/versions/.

See small changelog at https://zope.readthedocs.io/en/2.13/CHANGES.html#id1
and the versions file: http://dist.plone.org/versions/zope-2-13-27-versions.cfg

diff with 2.13.27 is:
```
1a2
> Zope2 = 2.13.28
32d32
< Zope2 = 2.13.27
43a44
> six = 1.11.0
```

mauritsvanrees:
And I see we have Zope2 2.13.29 now.  Diff between that and our 2.13.27 versions file:

```
1a2
> Zope2 = 2.13.29
24,25c25,26
< Products.ZCatalog = 2.13.29
< Pygments = 2.2.0
---
> Products.ZCatalog = 2.13.30
> Pygments = 2.3.1
32d32
< Zope2 = 2.13.27
34a35
> filelock = 3.0.10
37,39c38,40
< mr.developer = 1.38
< pluggy = 0.6.0
< py = 1.5.2
---
> mr.developer = 1.34
> pluggy = 0.8.1
> py = 1.7.0
43a45
> six = 1.12.0
45c47,48
< tox = 2.9.1
---
> toml = 0.10.0
> tox = 3.7.0
47c50
< z3c.checkversions = 0.5
---
> z3c.checkversions = 1.1
```

The only really relevant changes are that we should use these pins now:

```
Zope2 = 2.13.29
Products.ZCatalog = 2.13.30
```

I am making a PR to add those to cored 4.3.x. We could still do a separate zope2 versions file. Slightly cleaner, but I don't much mind.

mauritsvanrees:
I have access to dist.plone.org since yesterday, so I compiled versions files for Zope2 2.13.28/29/30 and put them in https://dist.plone.org/versions/
28 and 29 will not be used actually, but I will create PRs for Plone 4.3 and 5.1 to use our 30 file.

PR #463, opened 2018-04-25
Use Zope2 2.13.28 [4.3]

mauritsvanrees:
This can be done differently, by pointing to a new Zope2 versions file, see #462, but for now it is at least good to test the new Zope version.


jensens:
@mauritsvanrees do i get it right, that this should be merged?

mauritsvanrees:
Ah, can be merged yes. I will do so now.

PR #466, opened 2018-04-26
Pin Products.ZCatalog back to 4.0.1.

mauritsvanrees:
GenericSetup has test failures with Products.ZCatalog 4.1
See https://github.com/zopefoundation/Products.GenericSetup/issues/62

Since we are already using ZCatalog 4.x, updated the source branch from 3.0.x to master.


mauritsvanrees:
This fixes the GenericSetup tests, and leaves only an ATContentTypes failure, but that one got fixed already in https://github.com/plone/Products.ATContentTypes/pull/55.
I will merge and it should be green then finally. Phew!

PR #467, opened 2018-04-28
5.2 more py3 branches

ale-rt:


ale-rt:
Current status:
image
Note the js error because caused by merging bytes with text...
Should be easily fixable

ale-rt:
Current status
image
Note the js error caused by merging bytes with string...
It is easily fixable but for today is good enough!

PR #469, opened 2018-05-04
Update build tools [5.1]

ale-rt:


mauritsvanrees:
Jenkins says:

    Module pip._vendor.webencodings.tests does not define any tests

I have rebased this PR and have added `pip` to the excluded packages. Let's see if that helps.

mauritsvanrees:
That helped. :-) Exactly one less test was run, so sounds good.

PR #470, opened 2018-05-04
Let tox share the checkout

ale-rt:
In this way we can test at once all thye python versions

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -13,18 +13,16 @@ extras =
     test
 
 commands =
-    {envbindir}/buildout -c {toxinidir}/py3.cfg buildout:directory={envdir} bootstrap
-    {envbindir}/buildout -c {toxinidir}/py3.cfg buildout:directory={envdir} annotate
-    {envbindir}/buildout -c {toxinidir}/py3.cfg buildout:directory={envdir} install test
-    {envbindir}/buildout -c {toxinidir}/py3.cfg buildout:directory={envdir}
+    {envbindir}/buildout -c {toxinidir}/py3.cfg buildout:directory={envdir} buildout:sources-dir={toxinidir}/src bootstrap
+    {envbindir}/buildout -c {toxinidir}/py3.cfg buildout:directory={envdir} buildout:sources-dir={toxinidir}/src annotate
+    {envbindir}/buildout -c {toxinidir}/py3.cfg buildout:directory={envdir} buildout:sources-dir={toxinidir}/src install test
+    {envbindir}/buildout -c {toxinidir}/py3.cfg buildout:directory={envdir} buildout:sources-dir={toxinidir}/src
     coverage run {envbindir}/test -v1 --auto-color {posargs}
 
 setenv =
     COVERAGE_FILE=.coverage.{envname}
 
-deps =
-    setuptools==39.0.1
-    zc.buildout==2.11.2
+deps = -r{toxinidir}/requirements.txt
     coverage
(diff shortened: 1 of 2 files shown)


wsgi.cfg
```diff
@@ -19,19 +19,19 @@ environment-vars =
 
 [uwsgi_ini]
 recipe = collective.recipe.template[genshi]:genshi
-input = templates/uwsgi.ini
+input = ${buildout:directory}/templates/uwsgi.ini
```
loechel, line 22: good catch, that this necessary

tox.ini
```diff
@@ -13,36 +13,32 @@ extras =
...
+    {envbindir}/buildout -c {toxinidir}/py3.cfg buildout:directory={envdir} buildout:sources-dir={toxinidir}/src
     coverage run {envbindir}/test -v1 --auto-color {posargs}
 
 setenv =
     COVERAGE_FILE=.coverage.{envname}
 
-deps =
-    setuptools==39.0.1
-    zc.buildout==2.11.2
+deps = -r{toxinidir}/requirements.txt
```
loechel, line 25: yes in the current situation the better way on pinning the requirements, in the long run it should be:

    deps = setuptools zc.buildout -c https://dist.plone.org/bootstrap-buildout-constraints.txt

tox.ini
```diff
@@ -13,36 +13,32 @@ extras =
     test
 
 commands =
-    {envbindir}/buildout -c {toxinidir}/py3.cfg buildout:directory={envdir} bootstrap
-    {envbindir}/buildout -c {toxinidir}/py3.cfg buildout:directory={envdir} annotate
-    {envbindir}/buildout -c {toxinidir}/py3.cfg buildout:directory={envdir} install test
-    {envbindir}/buildout -c {toxinidir}/py3.cfg buildout:directory={envdir}
+    {envbindir}/buildout -c {toxinidir}/py3.cfg buildout:directory={envdir} buildout:sources-dir={toxinidir}/src bootstrap
```
loechel, line 16: yes that makes sense, due to buildout internals to place sources-dir relative to buildout:directory

tox.ini
```diff
@@ -13,36 +13,32 @@ extras =
...
+    {envbindir}/buildout -c {toxinidir}/py3.cfg buildout:directory={envdir} buildout:sources-dir={toxinidir}/src annotate
+    {envbindir}/buildout -c {toxinidir}/py3.cfg buildout:directory={envdir} buildout:sources-dir={toxinidir}/src install test
+    {envbindir}/buildout -c {toxinidir}/py3.cfg buildout:directory={envdir} buildout:sources-dir={toxinidir}/src
     coverage run {envbindir}/test -v1 --auto-color {posargs}
```
loechel, line 20: at the moment plone.testing and plone.app.testing are not python 3 compatible, 
@icemac has started to port them, but that is not finished or merged, so that will propably fail most of the time, now.

For the long run that is what we need.

loechel (approved):
Good improvement

tox.ini
ale-rt: Thanks @loechel!
I am anyway using the test command to fix some packages (e.g. plone.supermodel, that do not require plone.testing).

icemac: @loechel `plone.testing` has been completely ported to Python 3, see plone/plone.testing#48, but it is not yet merged. It should be possible to use the branch of this PR here.

PR #471, opened 2018-05-08
Include test section but without Archetypes; include chameleon fix

davisagli:
With these changes we can start running tests on Python 3.


ale-rt (approved):
Indeed!

PR #474, opened 2018-05-15
Remove plips/cmf-merge.cfg plips/plip-1334-PlonePAS.cfg

mauritsvanrees:
No longer actively used.
They reference outdated CMF clones within the plone organisation.


mauritsvanrees:
I could start a Jenkins job, but nothing is using this.

PR #479, opened 2018-05-17
Pin Chameleon in such a way that mr.developer can unpin it

ale-rt:
Fixes https://github.com/plone/Products.CMFPlone/issues/2421


ale-rt:
@pbauer can you confirm this fixes plone/Products.CMFPlone#2421?

PR #480, opened 2018-05-18
Use the Python 3 branches of plone.testing and plone.app.testing.

icemac:


jensens:
The PR job can test it as well. And we moreover want to merge those two into its masters.

PR #481, opened 2018-05-23
Update to Zope 4 beta 5

gforcada:


gforcada:
@pbauer @plone/framework-team although there are 2 test failures (on CMFPlone checking for valid ids), at least tests run, not like in 5.2 jenkins jobs that buildout can not finish due to a dependency not being satisfied (Products.PluggableAuth... requires Zope 4.0b5 while we are pinning to b4).

So given that, I'm merging it.

PR #489, opened 2018-06-20
Remove jenkins cfg [5.2]

gforcada:
There was really nothing left useful there.

There will be pull requests also for 5.1 and 5.0. As the pull request job works across all these branches (I kept jenkins.cfg it on 4.3 as it has some more extra bits) I prefer to do clean up on all branches rather than keeping it only on 5.2 and then, surprise surprise, something breaks. :sweat_smile:


jensens:
@gforcada unification and simplification in buildout is always good. Please go on and merge them yourself as soon as you think its the time. I suppose it need to be synchronised with the test-setup?

PR #492, opened 2018-06-20
Remove more outdated cfgs

gforcada:
@jensens is it ok to remove the ``zeoserver.cfg`` one? :thinking: 

For all the others, I guess they are totally fine, some of them have been untouched for more than 8 years :sweat_smile:


jensens:
Ok for coredev.

PR #495, opened 2018-06-23
Checkout DX container for P5.1

jaroel:
do not merge

### DIFF
--- a/checkouts.cfg
+++ b/checkouts.cfg
@@ -1,6 +1,7 @@
 [buildout]
 always-checkout = force
 auto-checkout =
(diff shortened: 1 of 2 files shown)


davisagli:
I'd suggest making a plip cfg file in the 5.2 branch rather than a separate branch. That way we don't have to mess around with rebasing/merging while testing and reviewing the changes.

jaroel:
I've added a plip.cfg + jenkins plip job: https://jenkins.plone.org/view/PLIPs/job/plip-dx-siteroot/

PR #496, opened 2018-06-25
Remove old pins in some cases causing downgrades (zodbpickle, persist…

sunew:
…ent, BTrees) from Zope 4.0b5 (I guess zope version was incremented and pins were not checked and removed)


jensens (approved):
All pins in the official Zope4 release cfg should not show up here. Downgrades are also not what we want. 
So, removing them to avoid downgrade LGTM.

PR #504, opened 2018-07-12
Try to fix 5.2 branch

gforcada:
Since a few builds one of the jenkins jobs targetting 5.2 branch is failing, my guess is that https://github.com/plone/plone.indexer/pull/6 was the problem as it was merged although the 5.2 jenkins job pull request failed, was there a good reason to ignore it? @jensens 

Let's see if, indeed, removing ``plone.indexer`` fixes the problem.

### DIFF
--- a/checkouts.cfg
+++ b/checkouts.cfg
@@ -49,4 +49,4 @@ auto-checkout =
     plone.formwidget.namedfile
     plone.tiles
     Products.ExtendedPathIndex
-    plone.indexer
+


jensens:
At time of merging it was green IIRC. Also, there were other PRs merged together with plone.indexer.

I think a side effect came in with ``plone.restapi``, but no idea why it did not hit the PR...
here https://jenkins.plone.org/job/pull-request-5.2/962/testReport/plone.restapi.tests.test_search/TestSearchFunctional/test_full_metadata_retrieval/

The problem: Before ``exclude_from_nav`` metadata was accidentally acquired. Now it is not and the returned value is ``None`` instead of the previously acquired ``False``.

Question here is, 
- do we want to accept the new behavior, which is telling us there was no explicit value set? 
- or do we want to register an indexer returning false on items w/o the attribute ``exclude_from_nav``?

jensens:
should be solved with https://github.com/plone/plone.restapi/pull/580

PR #516, opened 2018-09-26
Remove old generate_gruntfile.py

sunew:
these days gruntfile generation takes place in Products/CMFPlone/_scripts
with ./bin/plone-compile-resources

@ebrehault did I get this right?


jensens:
this is correct.

PR #517, opened 2018-10-02
Drop interlude

gforcada:
Seems that no package depends on it and is only added on our test suite
as it is specifically added there.

The version pin is left in case someone is still using it, to not get an
unexpected version bump.


gforcada:
Last changes on that package date from 2014... https://github.com/collective/interlude/commits/master

mauritsvanrees (approved):

PR #523, opened 2018-10-03
Pin plone.app.robotframework to a branch for automatic port selection.

Rotonen:
After discussing with @jensens, we've decided to try my branch out for automatically picking ports for Robot test layers to allow to run our layers in parallel.

https://github.com/plone/plone.app.robotframework/pull/86


Rotonen:
Obsoleted by getting `plone.app.robotframework` merged.

PR #538, opened 2018-10-18
allow to add eggs from part buildout

pbauer:
also add devtool-eggs to wsgi

### DIFF
--- a/buildout-py2.cfg
+++ b/buildout-py2.cfg
@@ -31,6 +31,7 @@ user = ${buildout:plone-user}
 http-address = 8080
 eggs =
     Plone
+    ${buildout:eggs}
     ${buildout:devtool-eggs}
 deprecation-warnings = on
 environment-vars =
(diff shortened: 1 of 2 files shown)


jensens (commented):
Nope. Please can we get rid of ${buildout:eggs} at all? They are downloaded and built before the parts and so control of execution is a bit difficult which may lead in edge cases to nasty, time consuming  problems. Which I had in past and who knows what happens in future?

Tell me a single valid reason when they are needed.

+1 for adding devtool-eggs to wsgi.

pbauer:
Not sure. Why are they downloaded and built before the parts? It is just something I'd have expected to work. Otherwise I"d have to add eggs to all part where I want them like this in my `local.cfg`:

```
[wsgi]
eggs +=
    some.egg

[instance]
eggs +=
    some.egg

[instance-archetypes]
eggs +=
    some.egg
```
That seems like a lot of text for simply adding a additional egg to my local coredev for testing something. I did somethink like this to fix the demo: https://github.com/collective/demo.plone.de/commit/39b83862a1e62be4a85956920db156947901d3ff

Or I could abuse `devtool-eggs` for that... 

By the way: Did you check if the plip-job setup still works after the changes you did?

jensens:
replaced by #539

PR #541, opened 2018-10-23
Update versions.cfg

gforcada:
Test the new persistent release

### DIFF
--- a/versions.cfg
+++ b/versions.cfg
@@ -67,7 +67,7 @@ msgpack = 0.5.6
 Products.ZCatalog = 4.2
 # until https://github.com/plone/Products.CMFPlone/issues/2455 is resolved
 ZODB = 5.3.0
-
+persistent = 4.4.3
 # Plone dependencies on other Zope packages not part of the release
 Products.ExternalMethod = 4.0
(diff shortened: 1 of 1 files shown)


icemac:
At least the `safe_formatter` tests need zopefoundation/Zope#387.

Rotonen:
Is that oid correct here?

https://jenkins.plone.org/job/pull-request-5.2-3.7/20/testReport/junit/Products.CMFQuickInstallerTool/tests/install_txt/

icemac:
@Rotonen wrote:
> Is that oid correct here?
> https://jenkins.plone.org/job/pull-request-5.2-3.7/20/testReport/junit/Products.CMFQuickInstallerTool/tests/install_txt/

`persistent` 4.4.3 adds the module name before the class name in the `repr`.
https://github.com/zopefoundation/Zope/pull/387 aims to restore the `repr` which was used before the changes in `persistent` 4.4.0. Could you try that branch? (If it does not help you have to add the new `OFS.SimpleItem.PathReprProvider` as first base class so its `__repr__` implementation is used. (It is only important to have it in the MRO before `persistent.Persistent`.

PR #542, opened 2018-10-26
Test Zope 4.0b7 before release

pbauer:


gforcada:
@pbauer weren't supposed to be disabled on 5.2 jobs the AT packages tests? :thinking: some of the failures here are due to that, but not only, there are some valid ``__repr__`` errors still.

PR #547, opened 2018-11-03
Fix PLIP base config to work with new WSGI config.

thet:


davisagli:
Note that this should become unnecessary if my changes in https://github.com/plone/plone.recipe.zope2instance/pull/52 get merged.

agitator:
shouldn't plipbase.cfg extend from buildout-py2.cfg or buildout-py3.cfg?

thet:
@agitator - @gforcada is working on a solution on that. see: https://github.com/plone/buildout.coredev/issues/549

jensens:
I think it does not hurt if we merge, even if it becomes obsolete soon.

PR #550, opened 2018-11-05
Replace version pin on slimit with calmjs.parse

metatoaster:
As required by plone/Products.CMFPlone#2616.

### DIFF
--- a/versions.cfg
+++ b/versions.cfg
@@ -85,6 +85,7 @@ zope.password = 4.3.1
 
 ##############################################################################
 # External dependencies
+calmjs.parse = 1.1.3
 cssselect = 1.0.3
 decorator = 4.3.0
 enum34 = 1.1.6
@@ -111,12 +112,11 @@ pytz = 2018.5
 repoze.xmliter = 0.6
 roman = 3.0
 simplejson = 3.16.0
-slimit = 0.8.2+plone
 urllib3 = 1.23
 webob = 1.8.2
 
 # Special pins, see annotations
-ply = 3.4
+ply = 3.11
 unidecode = 0.04.1
 
 ##############################################################################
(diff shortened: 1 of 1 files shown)


gforcada:
@metatoaster this needs to be merged on branch **5.2**, it *might* be backported to 5.1, but so far it should be only for 5.2 branch

many thanks for unlocking this blocker!! :bowing_man:

metatoaster:
Whoops, I was wondering why 5.1 was showing up for the CI. Thanks for pointing out this obvious mistake, and this looks more like it now.

metatoaster:
Ah, I only just noticed the notes on the special pinning on the `ply` package.  For the `calmjs.parse` wheel releases I can only pre-generate the *tab module files with only the supported released versions of `ply` thus far, so pinning to 3.11 (the current/latest released version as of now) is safe.  Should I also update the rest of the `versions.cfg` with this information?

gforcada:
@metatoaster yes, if that makes jenkins go green, sure.

There should be a pull request on cmfplone as well, or is that one already merged? :thinking:

metatoaster:
@gforcada I updated the note and removed that extra "temporary" redefinition as that is now the definitive one (but kept `ply` at the "Special pin" section).  Neither this nor my change on CMFPlone has been merged yet.  I also took the opportunity to update the version pin as sadly I failed to account for an irregular syntax case with the JavaScript language grammar which I just released a fixed (via calmjs.parse-1.1.3). Should hopefully be the last one I missed (which should have been eradicated the first time I encountered that for the `function` keyword, but hindsight, etc...)

pbauer:
Since https://github.com/plone/Products.CMFPlone/pull/2616 works now I'll merge this.

PR #553, opened 2018-11-10
Include plone app locales translations

anuyens:
Because some msgid was removed from 5.1 to 5.2, to ensure backward compatibility, I place all removed msgid in Python file (backward_compat_51.py) on plone.app.locales (the checkin follow on a branch called new_msgid_5.2).
For this, we must search also in plone.app.locales
Thank's


jensens:
@anuyens  this PR looks very broken. I propose to close it and you re-create it.

anuyens:
Hello jensens, ok to close and recreate a new PR (I do that this week-end)

PR #555, opened 2018-11-14
Use buildout conditions

jensens:
Have again one `buildout.cfg` rule them all. This is possible using [conditional configuration sections](http://docs.buildout.org/en/latest/topics/variables-extending-and-substitutions.html#conditional-configuration-sections). This also helps with our plip-jobs

@gforcada I suppose the change here needs some changes in our Jenkins setup? 

cc @pbauer @davisagli @ale-rt


gforcada:
@jensens yes, please do not merge it, let me do it, as jenkins.plone.org jobs would need to be adapted.

gforcada:
jenkins.plone.org adapted, let's see if it flies or sinks :sweat_smile:

PR #556, opened 2018-11-15
Move conditionals

jensens:
Move the conditional sections to the files they belong to. It makes no sense to have them all in `buildout.cfg`.


gforcada:
You could as well remove buildout.cfg completely as it is just extending core.cfg ...

jensens:
I had the same thought. I am not sure why we have/had it this way.

mauritsvanrees:
I now no longer have a `bin/develop` script. I created a fresh virtualenv with python 2.7.

```
$ bin/buildout 
Unused options for buildout: 'always-checkout' 'auto-checkout' 'config-directory' 'template-directory' 'zope-directory'.
Updating test.
Updating robot.
Updating instance.
Updating instance-archetypes.
Updating ploneversioncheck.
Versions had to be automatically picked.
[versions]
pyrepl = 0.8.4
$ ls bin
activate                buildout                easy_install            instance-archetypes     pip2.7                  python                  ride                    slimit
activate.csh            diazocompiler           easy_install-2.7        libdoc                  plone-compile-resources python-config           robot                   test
activate.fish           diazopreprocessor       icalendar               pip                     pybabel                 python2                 robot-debug             versioncheck
activate_this.py        diazorun                instance                pip2                    pybot                   python2.7               robot-server            wheel
```

Ah, `bin/buildout annotate` tells me `extensions` only has `plone.versioncheck`. So with this:

```
[buildout]
extensions +=
    mr.developer
[buildout:python27]
extensions +=
    plone.versioncheck
```

the result is that only `extensions += plone.versioncheck` is kept on Python 2.7.
Jens, can you have a look?

jensens:
🤔 Odd. Looks like conditionals behave different than I suspected. I 'll habe a look at it tomorrow.

PR #557, opened 2018-11-15
Do not allow packages without a pinned version

gforcada:
New packages are always welcome,
but knowing why they got introduced, maybe by mistake, 
could help identify and discuss the introduction of third party dependencies.

### DIFF
--- a/core.cfg
+++ b/core.cfg
@@ -32,6 +32,7 @@ extensions =
     mr.developer
 
 show-picked-versions = true
+allow-picked-versions = false
 
 # define a template directory here in order to override it later in custom
 template-directory = ${buildout:directory}/templates


gforcada:
As a matter of an example, it took me quite some hours to dig out which package was introducing ``robotframework-python3`` dependency (spoiler: robotsuite :smile: ).

Or just now that I pinned the last new set of packages (related to pdbpp mostly) that were unpinned.

mauritsvanrees (dismissed):
Agreed on setting `allow-picked-versions = false`.
But I would keep the `show-picked-versions = true` setting. That combination is what I have in most client buildouts.
Say you update some package pin to a newer version and this introduces five new dependencies. Without `show-picked-versions` you would need to run buildout five times and add one new version each time, before you have all extra dependencies. With `show-picked-versions = true` you only need to temporarily set `allow-picked-versions` to true and then buildout will finish and give you a list of all new package pins for you to copy.

jensens:
I agree with @mauritsvanrees

mauritsvanrees (approved):

mauritsvanrees:
Well, approved, but when I try it out, I already get:

```
Error: Picked: pyrepl = 0.8.4
```

And then my `bin/buildout` script is suddenly gone, which is weird but should not have anything to do with this...

gforcada:
I pinned this one, maybe it is a python 2 only dependency? :thinking: I didn't see it on python 3...

Anyway, pinned, and at least locally it works :tm:

jensens (approved):

mauritsvanrees:
So it could be that locally with Python 3 everything works, and on Python 2 the buildout doesn't even finish. Well, that shouldn't happen very often, and if the change is done in a PR then Jenkins will catch it.

mauritsvanrees:
I have rebased this branch on 5.2, force pushed it, and started another batch of Jenkins jobs.

mauritsvanrees:
All tests pass, except known unstable tests. I will merge.

PR #559, opened 2018-12-14
Upgrade to Zope 4.0b8

pbauer:


pbauer:
For Zope 4.0b7 we used our own http://dist.plone.org/versions/zope-4-0-b7-versions.cfg which combined versions.cfg and versions-prod.cfg of Zope. 

We should do the same for https://raw.githubusercontent.com/zopefoundation/Zope/4.0b8/versions.cfg and https://raw.githubusercontent.com/zopefoundation/Zope/4.0b8/versions-prod.cfg to be independent of github.

pbauer:
All failing robot-tests (`test_edit_user_schema.robot`) pass locally and the feature also works in manual testing. I'm calling 'Bad bad Robot!' on that and would like to merge.

gforcada:
Maybe the workers were too stressed out and that's why they failed? :man_shrugging:

gforcada:
Scratch the previous comment, seems that indeed, robot tests on chrome have been broken since a few days already :confounded:

PR #562, opened 2019-02-04
Fix parts for Python 2.7

pbauer:
packages, zopepy and other parts were missing in py2. Fixes https://github.com/plone/Products.CMFPlone/issues/2711

### DIFF
--- a/core.cfg
+++ b/core.cfg
@@ -6,7 +6,7 @@ extends =
     versions.cfg
     tests.cfg
 
-parts +=
+parts =
     instance
     zopepy
     packages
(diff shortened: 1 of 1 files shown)


core.cfg
```diff
@@ -6,7 +6,7 @@ extends =
     versions.cfg
     tests.cfg
 
-parts +=
+parts =
```
thet, line 9: @pbauer the inclusion of tests.cfg above is now almost obsolete - the parts defined in tests.cfg (test, robot) are overwritten by this section.
Is that intentional or should we re-add test + robot here?

PR #563, opened 2019-02-07
Add plone.staticresources to the test eggs

ale-rt:
Refs https://github.com/plone/Products.CMFPlone/issues/1653


ale-rt:
@gforcada this will fix the fact that in this PR https://github.com/plone/plone.staticresources/pull/8 the jenkins jobs are not triggered or do we need to do something else?

gforcada:
Probably as it is a *new* package, the hooks were not installed on it...

PR #564, opened 2019-02-10
Update to Zope 4.0b9

pbauer:


pbauer:
None of the test-failures are locally reproducable. I'll restart the testrun.

pbauer:
Darn. The three test 
`test_navroot_params_on_404_widget_in_path`, `test_theme_for_404` and `test_theme_params_on_404` from `plone.app.theming` fail when running all tests locally. When running only these three or all tests of `plone.app.theming` they pass. So this seems like a new test-isolation-issue.
The task is now to find out which other tests are responsible for these failures.

jensens:

jensens:
this one should be more verbose (output of the page content)
https://jenkins.plone.org/job/pull-request-5.2-3.7/354/parameters/

pbauer:
The failure happens only when running the layers `zope.testrunner.layer.UnitTests` and `plone.app.theming.testing.Theming:Functional`: 
```
./bin/test --layer 'zope.testrunner.layer.UnitTests' --layer 'plone.app.theming.testing.Theming:Functional'
```

pbauer:
More specifically it is the test `zope.rst` in `plone.testing` that causes the failure in `plone.app.theming`:

```
./bin/test -s plone.app.theming -s plone.testing -t test_theme_for_404 -t zope.rst
```

pbauer:
And even more specifically: The test in plone.app.theming only fails after a setup of the layer ```plone.testing.zope.WSGIServer``` with 

```
>>> runner.setup_layer(options, zope.WSGI_SERVER, setupLayers)
```

pbauer:
@icemac does that maybe ring a bell?

icemac:
@pbauer Sorry, no, I have no idea why these tests are failing now but not using Zope 4.0b8.

pbauer:
The test fails since this commit in Zope: https://github.com/zopefoundation/Zope/commit/ba72a30cf6332fc060e38e3e1849be407180b75a to fix https://github.com/zopefoundation/Zope/issues/413

Rotonen:
What's that Zope startup handler change about?

pbauer:
I finally found it in my heart to reproduce what the test actually tests and  - oh my!

The test checks if 404's are themed, as they should be. 

But now when no valid result is found for the current request you see a blank page. The default 404-page (error_message.pt) seems only empty because the theme is not applied (see below). 

You can easily reproduce it when you open http://localhost:8080/Plone/foo (when there is no content with the id  `foo` and you use this branch). 

Before the change in Zope `HTTPRequest.retry_max_count` was `0` and is now `3`. When I switch it back to `0` it works again.

I dug a little deeper. 

- The `ExceptionView` of Plone is called as usual and the template-rendering works fine.

- `applyTransformOnFailure` (in `plone.transformchain.zpublisher`) adapts `IPubBeforeAbort` and aborts because `event.retry` is now True. Thus `applyTransformOnSuccess` is not called and the response-body is not modified by the registered transforms. The now missing transform `plone.app.theming.transform` results in a response-body without the theme.

- The result only looks empty because the `<script>`-tags are self-closing:
`<script type="text/javascript" src="..." />` which is invalid if the header specifies the mimetype as `text/html` instead of the valid `<script type="text/javascript" src="..."></script>`

What really tripped me first up was that when I inspected the source in the Firefox developer toolbar the html seemed empty. Only when I finally used right-click-view-source I saw the html (with the closing slash in <script/> colored red).

Back to the real problem: 

In `transaction_pubevents` the call `request.supports_retry()` now is true and it never was before...

in Zope 4 it is https://github.com/zopefoundation/Zope/blob/master/src/ZPublisher/WSGIPublisher.py#L164

In Zope 2 it is https://github.com/zopefoundation/Zope/blob/2.13/src/ZPublisher/Publish.py#L191

Maybe @icemac or @dataflake can tell me how to fix this.

dataflake:
https://github.com/zopefoundation/Zope/pull/473 was done to actually make the retry-on-conflict work again. Before it was completely broken.

pbauer:
In Zope2 `retry` was only `True` when a `Retry` was raised and thus the request would be retried. In all other cases `retry` was `False`. `request.supports_retry()` was only called to determine if the request should be retried.

I do not think we should use `request.supports_retry()` when notifying `PubBeforeAbort` and `PubFailure`. The flag is rather meant to signal if the current request will be retried or not.

pbauer:

gforcada:
@pbauer hats off on your debugging session! :rocket:

Rotonen:
So is there still a WSGI layer setup issue that the setup / teardown of the WSGI publisher is somehow silly and leaks?

pbauer:
@Rotonen good question. 

Since `retry_max_count` is set on `ZPublisher.HTTPRequest.HTTPRequest` each new request in the whole test-run has the value `3`. It seems that the tests `layer.rst` and `zope.rst` of plone.testing are so far the only tests in Plone that set up `plone.testing.zope.WSGIServer`. Zope has some more tests but we do not run them when testing Plone.

I'm not sure if that qualifies as a test-isolation-breach since this setting is the default that is also used when actually running Plone. 

Maybe we should instead write a couple more tests that actually use that layer.

dataflake:
Your PR zopefoundation/Zope#502 is now merged int o Zope.

pbauer:

jensens:
So, we skip Zope 4.0b9 and wait for Zope 4.0b10, but go with Zope 4.0b8 for the Plone 5.2b1 release.

PR #570, opened 2019-02-21
[5.1] update mosaic versions

petschki:


petschki:

petschki:
@jensens sorry but this has to be reverted because I've updated `plone.subrequest = 1.9.0` and this unfortunately does not pass in Plone 5.1 tests ...

jensens:
does it work with 1.8.6?

jensens:
reverted subrequest bump only for now

petschki:
@jensens yes just tested with 1.8.6 and works ...I've removed the pin in my latest plone.app.mosaic PR https://github.com/plone/plone.app.mosaic/pull/441

PR #571, opened 2019-02-21
use plone.app.mosaic 2.1.x versionset for Plone 5.1

petschki:
as discussed here: https://github.com/plone/plone.app.mosaic/pull/441


jensens (approved):
please just merge if green

PR #576, opened 2019-03-27
Always run virtualenv with python 2.7 in bootstrap.sh for 5.1

fredvd:
The latest virtualenv by default installs a python 3.x/3.7 environment without -p option.
Theoretically python2 would be enough, but anything below 2.7 is not supported anyway.


mauritsvanrees:

mauritsvanrees (approved):
LGTM

jensens (approved):

PR #579, opened 2019-04-10
Let flake8 ignore W503: Line break occurred before a binary operator.

mauritsvanrees:
With line breaks and binary operators you will *always* hit either this warning:

  W503: Line break occurred before a binary operator

or this warning:

  W504: Line break occurred after a binary operator

The pep8 guidelines were changed to recommend the line break before.
See https://lintlyci.github.io/Flake8Rules/rules/W503.html

So we ignore W503.

See the code analysis report that I just got for [`plone.api`](https://github.com/plone/plone.api/commit/b5eaefab97162e8a712d39a1b0c95ef98043c653).

### DIFF
--- a/experimental/qa.cfg
+++ b/experimental/qa.cfg
@@ -17,6 +17,12 @@ pre-commit-hook = False
 jenkins = True
 check-manifest = True
 clean-lines = True
+# With line breaks and binary operators you will always hit one of these:
+# W503: Line break occurred before a binary operator
+# W504: Line break occurred after a binary operator
+# The pep8 guidelines were changed to recommend the line break before.
+# See https://lintlyci.github.io/Flake8Rules/rules/W503.html
+flake8-ignore = W503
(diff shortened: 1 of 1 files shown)


jensens:
One puzzle tile of https://github.com/plone/Products.CMFPlone/issues/2754

gforcada:
@mauritsvanrees thanks, as @jensens as soon as we have `black` in, we have to fine tune `flake8` to not complain about formatting, but only about possible bugs and idiomatic coding only.

As the PLIP got accepted, maybe is pointless to keep generating these reports. I will disable the jenkins job. The next time it will be enabled it should be with both `black` and fine tuned `flake8` :+1:

gforcada (approved):

mauritsvanrees:
Disabling it may make sense yes. Sadly not a lot of people seem compelled to do anything with these reports, often me included, though I run some checks locally, and try to at least not make the flake8 complaints worse. 
autopep8/black FTW!

PR #580, opened 2019-04-18
Add volto config

pbauer:
- Add CORS-Headers
- Enable folderish default-types
- Install plone.restapi by default
- Enable tiles-behavior


jensens:
Since Volto lives as a PLIP, the usual way is to add this to ``./plips/plip-xxxx-volto.cfg`` - or in the case that it is difficult to have it in a subfolder to ``./plip-xxxx-volto.cfg``.

PR #581, opened 2019-04-18
Update versions

pbauer:


pbauer:
The tests took pretty long, so I'm running them again on node 1. I guess node 2 simply was busy with something else but I'd like to be sure we're good before merging.

pbauer:
Testruns are ok on Node 1. Node 2 is extremely slow.

PR #582, opened 2019-04-30
Test Zope master

pbauer:
Do not merge! This is only here to prevent surprises before the Plone 5.2 final release


jensens:
Maybe a (manual?) job testing Plone coredev together with of Zope/CM "coredev" in Jenkins would be a good investment to avoid some surprises?

pbauer:
replaced by https://github.com/plone/buildout.coredev/pull/589

PR #587, opened 2019-05-21
Checkout z3c.form, pin zope.schema 4.6.0

mauritsvanrees:
Both coredev 5.1 and 5.2 use the 3.x branch of z3c.form, which has changes, but only 5.2 has it in the checkouts. So we add z3c.form to checkouts.cfg in 5.1.  This needs zope.schema 4.6.0, so we update the pin.


mauritsvanrees:

mauritsvanrees:
This gives a [test failure](https://jenkins.plone.org/job/pull-request-5.1/47/testReport/) in plone.app.users and two in z3c.form.
It may be best to create a 3.6.x branch of z3c.form before the changes from October 2018, and use that branch in coredev 5.1. Then we don't have to update the zope.schema pin, which may also cause test failures in add-ons.

mauritsvanrees:
Done.  Closing this PR.

PR #601, opened 2019-08-30
Bump lxml version

gforcada:
Previous version, was making jenkins (on python 3.8b4) fail with a `Segmentation fault`, which was actually hiding the fact that lxml was somehow not built properly


gforcada:

mauritsvanrees:
Thank you!!! 👍🏻

PR #604, opened 2019-09-18
generate constraints.txt

jensens:
For later use with pip like `pip install -c constraints.txt Plone Paste` 

After above works, for next release it would be great to have it released to dist.plone.org, 
@esteele What do you think?

This can be merged right away, it doe not hurt now.

### DIFF
--- a/core.cfg
+++ b/core.cfg
@@ -19,10 +19,16 @@ allow-hosts =
 find-links += http://dist.plone.org/thirdparty/
 
 extensions =
+    buildout.requirements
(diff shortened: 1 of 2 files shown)


jensens:

esteele (approved):

mauritsvanrees:
@jensens We should let git ignore `constraints.txt` then, right?

Or would you want to add this file to the repository? I think not: there are obviously differences in the constraints generated by Python 2 and 3.

mauritsvanrees:
I have added it too `.gitignore`.

PR #607, opened 2019-09-19
Upgrade to Zope2 2.13.29. [5.1.x]

mauritsvanrees:
Zope2 was at .27 (from http://dist.plone.org/versions/zope-2-13-27-versions.cfg)
Note that Plone 4.3 was already using .28.
See also https://github.com/plone/buildout.coredev/issues/462
And note that unlike 4.3 PR #606 we do not need to update ZCatalog, because 5.1 already uses a more recent version.

### DIFF
--- a/versions.cfg
+++ b/versions.cfg
@@ -8,6 +8,7 @@ extends = http://dist.plone.org/versions/zope-2-13-27-versions.cfg
 ################
 # Zope overrides
 docutils = 0.14
+Zope2 = 2.13.29
 # Get support for @security decorators and hotfixes
 AccessControl = 3.0.14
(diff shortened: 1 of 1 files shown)


mauritsvanrees:

ale-rt:
Sorry @mauritsvanrees, is this just temporary?
Because I see a few lines above:
```
extends = http://dist.plone.org/versions/zope-2-13-27-versions.cfg
```
I think it might be better to prepare a https://dist.plone.org/versions/zope-2-13-29-versions.cfg and replace that line.
But I am not really sure about the process here.

mauritsvanrees:
See issue #462.

PR #609, opened 2019-10-09
Revert Products.CMFCore to 2.2.10. [4.3]

mauritsvanrees:
In 2.2.11 collective.indexing was merged, so we never want that in Plone 4.3 or 5.0.
That merge should never have happened on the 2.2 branch, but that was in 2017, so we can't do anything about that.

In 2017 this already gave errors in Plone 5.0: https://github.com/plone/Products.CMFPlone/issues/2042
Now with 2.2.13, Jenkins is failing on 4.3: https://jenkins.plone.org/job/plone-4.3-python-2.7/26/

I worried that this may have gotten in last moment for Plone 4.3.19, but this is not the case.  It has the correct 2.2.10. Phew!

Note: we don't have any `versionannotations` section in the buildout, like we have on 5.x. I am not sure if that would have helped here.


mauritsvanrees:

mauritsvanrees:
Since this fixes the currently broken Jenkins, I will merge.

PR #614, opened 2019-11-04
Test fix for 1318 with plone.app.contenttypes branch

ericof:


jensens:
Well, you do not need to create a PR to test two branches together. Just create for each branch a PR and fill in both PR URLs into Jenkins ``Build with Parameters`` form, one per line.

PR #616, opened 2019-11-20
Update robot framework packages

ale-rt:


ale-rt:
I would focus on doing this in Plone 6

PR #618, opened 2019-12-02
Update Zope and Dependencies

pbauer:


requirements.txt
```diff
@@ -1,2 +1,2 @@
-setuptools==41.0.1
+setuptools==42.0.2
```
mauritsvanrees, line 1: I am a bit hesitant about this setuptools version. But it should be okay. It will give warnings, but that can be a good thing.
See this [buildout issue](https://github.com/buildout/buildout/issues/493#issuecomment-558334008)

mauritsvanrees (approved):
One inline comment about setuptools.
But seems fine, as long as Jenkins passes.
For the disk space error, I have reopened https://github.com/plone/jenkins.plone.org/issues/267

jensens (approved):
Great! May you provide the same for the 6.0 branch too?

pbauer:
@jensens Thanks for the reminder. I would have forgotten that.

pbauer:
Almost done. https://github.com/plone/plone.app.theming/pull/167 should fix the last remaining issue.

PR #648, opened 2020-03-24
Use python3 and not hardcoded python3.6 in bootstrap_py3

fredvd:
'which python3.6` is hardcoded in the boostrap_py3.sh script.

* we could use 3.7 or 3.8 but that might break
* we could be 'smart' and first try to detect 3.7 and then fall back to 3.6
* Is Python 3.8 already supported?
* lets stick to searching for python3, as the script is named for now

pros/cons?

(won't start a CI build until it is clear which python3 to pick)

### DIFF
--- a/bootstrap-py3.sh
+++ b/bootstrap-py3.sh
@@ -1,5 +1,5 @@
 #/bin/sh
-`which python3.6` -m venv .
+`which python3` -m venv .
 ./bin/pip install -r requirements.txt
 ./bin/buildout $*
 echo "run plone with: ./bin/instance fg"


jensens:
Sounds fine. But AFAIK the CI does not use the script. 3.8 is supported IIRC, 3.9 not (but its not final yet). We may want to add a check if the picked python version is in the range of supported versions (3.6-3.8).

mauritsvanrees:
I use 3.7 by default, but fine to use 3.8 if available.

This seems to work for me:

    `which python3.8 || which python3.7 || which python3.6` -mvenv .

thet:
I just added the exactly same change like @fredvd . I think his change is absolutely fine. Also @mauritsvanrees of course although the extra check for a specific python version isn't really necessary IMO and we'll forget to change it for sure once we do not support python3.6 anymore.

mauritsvanrees:
Fair enough. Whatever people have as python3 should usually be enough. And if they have a too new one, then they may detect problems early and report them, which is good.
I will merge.

mauritsvanrees (approved):

PR #649, opened 2020-03-26
Moved Products.validation pin and test-eggs to run on Python 3 as well.

mauritsvanrees:
In core, it is only used by Archetypes, so it will not end up in bin/instance.
But the dexterity add-on collective.easyform depends on it for its validators, for better or worse.
So would be good to check it on Python 3.

Locally with Python 3.6 its tests pass.

### DIFF
--- a/tests.cfg
+++ b/tests.cfg
@@ -113,6 +113,7 @@ test-eggs =
     Products.PluginRegistry
     Products.PortalTransforms
     Products.statusmessages
+    Products.validation
     Products.ZopeVersionControl
     repoze.xmliter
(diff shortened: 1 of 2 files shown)


mauritsvanrees:

mauritsvanrees:
Note: on Plone 6.0 we can still remove this, because then nothing in core depends on it.

mauritsvanrees:
Jenkins fails after this merge, but that is because of a [change in Zope](https://github.com/zopefoundation/Zope/pull/807) this morning. I have updated our Chameleon pin, and hope it will work.

PR #654, opened 2020-04-24
Extend Zope 2.13.30 versions file

mauritsvanrees:
This is a new versions file that I put online yesterday. See issue #462.
This gives us the chance to remove some own version pins. And we were still on Zope2 2.13.29 instead of 30.
Also, we were pinning `unittest2 = 1.1.0` in our `tests.cfg`, but the Zope versions file was extended later, so their version pin 0.5.1 won. So I got rid of our unittest2 pin. It was not pulled in by any of our packages anyway.

### DIFF
--- a/tests.cfg
+++ b/tests.cfg
@@ -163,7 +163,6 @@ launchpadlib = 1.10.5
 lazr.authentication = 0.1.3
 lazr.restfulclient = 0.13.5
 lazr.uri = 1.0.3
-linecache2 = 1.0.0
 lockfile = 0.12.2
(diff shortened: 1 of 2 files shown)


mauritsvanrees:

mauritsvanrees:
Two robot test failures that seem unrelated. I don't have them locally, but I do have one different robot failure. Let's try again.

mauritsvanrees:
Without any changes, now two different robot tests fail.
I guess Jenkins is just too busy, which can lead to errors like this.

PR #668, opened 2020-07-31
Fix windows

jensens:
Fix the buildout.coredev to run on windows. Adds Github Action to run buildout on windows.

PR #669, opened 2020-08-03
Fix Windows Plone 6

jensens:
more or less the same as #668 but for Plone 6.0 (w/o py 2.7 specific changes)


jensens:

PR #670, opened 2020-08-13
Base on Zope 4.5.1

jensens:
- Update to use newer Zope
- Update our own version pinnings
- Add a GH action to run plone.versioncheck and produce an artifact with its versions.html table.


jensens:

jensens:
Memo/Questions: 
- Pillow needs update
- robotframework and its libs are outdated, update?
- some git libs in test.cfg are outdated, action needed?

mauritsvanrees:
Jenkins on Python 3.6 and 3.7 has several failures with mail sending: `initial_value must be str or None, not bytes`. I wonder if that is because of the updated `Products.MailHost`. With Jenkins 2.7 I do not see this, and locally with 2.7 the CMFPlone tests pass. Solution for the moment would be to not update MailHost, and figure it out later.

On Python 2.7, there are failures for plone.app.folder, plone.dexterity, and plone.restapi. I can confirm the first two locally.

In general, at this moment I would prefer only the update to Zope 4.5.1, plus if needed removing a few pins if Zope has the same or newer. Then I can do a final 5.2.2 release without worrying that a last-minute minor package update causes harm.
I can also do that in a separate PR though.

mauritsvanrees:
I see mail-related tests passing when I downgrade Products.MailHost to 4.9.
Related are changes in [`zope.sendmail 5.1`](https://github.com/zopefoundation/zope.sendmail/pull/29), although with that version and the older Products.MailHost, the tests pass. So the problem is likely introduced in https://github.com/zopefoundation/Products.MailHost/pull/32.

mauritsvanrees:
Some of the other fails are likely because you removed the `zope.interface 5.0.2` pin, so we get version 5.1. That version is too new at this point.

mauritsvanrees:
Hm, going back to zope.interface 5.0.2 I still get this on Python 2.7:

```
Error in test test_dynamic_schema_refreshed_on_modify_schema_policy (plone.dexterity.tests.test_fti.TestFTIEvents)
  File "/Users/maurits/.pyenv/versions/2.7.17/lib/python2.7/unittest/case.py", line 329, in run
    testMethod()
  File "/Users/maurits/shared-eggs/cp27m/plone.dexterity-2.9.7-py2.7.egg/plone/dexterity/tests/test_fti.py", line 750, in test_dynamic_schema_refreshed_on_modify_schema_policy
    DexterityFTIModificationDescription('schema_policy', '')
  File "/Users/maurits/shared-eggs/cp27m/plone.dexterity-2.9.7-py2.7.egg/plone/dexterity/fti.py", line 561, in ftiModified
    syncSchema(model.schema, schema, overwrite=True, sync_bases=sync_bases)
  File "/Users/maurits/shared-eggs/cp27m/plone.supermodel-1.6.2-py2.7.egg/plone/supermodel/utils.py", line 338, in syncSchema
    dest.__bases__ = tuple(bases)
  File "/Users/maurits/shared-eggs/cp27m/zope.interface-5.0.2-py2.7-macosx-10.15-x86_64.egg/zope/interface/interface.py", line 404, in __setBases
    b.unsubscribe(self)
  File "/Users/maurits/shared-eggs/cp27m/zope.interface-5.0.2-py2.7-macosx-10.15-x86_64.egg/zope/interface/interface.py", line 391, in unsubscribe
    n = self._dependents[dependent]
  File "/Users/maurits/.pyenv/versions/2.7.17/lib/python2.7/weakref.py", line 363, in __getitem__
    return self.data[ref(key)]
KeyError: <weakref at 0x10ecd5230; to 'InterfaceClass' at 0x10ec39f80 (IBlank)>
```

mauritsvanrees:
Huh? On Python 2.7, with zope.interface 5.0.2, `bin/test -s plone.dexterity` passes, but in `bin/test -s plone.app.folder -s plone.dexterity` it fails with the error in my previous comment. So running the `plone.dexterity` tests in combination with `plone.app.folder` breaks one `plone.dexterity` test.
`plone.app.folder` is only for Python 2, so Python 3 does not have this problem.

I guess on Jenkins these two packages are in different groups, so maybe there is no problem.

Meanwhile, I have a branch with only Zope 4.5.1 as change (plus zest.releaser, just because I can). I will make a PR from that, and see what Jenkins says.

mauritsvanrees:
I have merged my smaller PR #671. If you want, you can rebase your changes on that. But some changes are needed, as indicated above.

Strangely, [Jenkins on Python 3.8](https://jenkins.plone.org/job/plone-5.2-python-3.8/701/testReport/plone.dexterity.tests.test_fti/TestFTIEvents/test_dynamic_schema_refreshed_on_modify_schema_policy/) shows the same `weakref` error in `plone.dexterity` for the past two builds, before merging my branch. Maybe we have a new flaky test...

jensens:
Well, I'll start over fresh and will focus on Zope 5.x/ Plone

PR #678, opened 2020-08-24
Do not use the emtpy icalendar ``test`` extra

thet:


mauritsvanrees:
5.0 is not updated anymore.

PR #682, opened 2020-09-29
Added the positive review of PLIP 1775

ale-rt:
Refs. plone/Products.CMFPlone#1775


ale-rt:
CC @plone/framework-team
Sorry @mauritsvanrees it took such an embarrassing long time.

jensens:
I am not sure if we need a PR for a review, but anyway, it does not hurt and it increases visibility!

PR #686, opened 2020-11-05
Update pep517 to version that works with latest check-manifest.

mauritsvanrees:
Updated check-manifest too.
It depends on 'build', added that too.
I ran into version conflicts with the easyform buildout, see https://github.com/collective/collective.easyform/pull/253

(At some point we may want to publish the `[versions]` part of `tests.cfg` on `dist.plone.org`. Or merge it with `versions.cfg` in coredev. That would have prevented this, because in `tests.cfg` we already pinned `check-manifest` to a version that works with the older `pep517`.)


mauritsvanrees:

jensens:
+1 for merging both files

PR #696, opened 2021-01-15
Remove Products.CMFFormController / portal_form_controller.

mauritsvanrees:
Remove the package from the buildout files.
Use branches of `CMFPlone` (https://github.com/plone/Products.CMFPlone/pull/3226) and `plone.app.upgrade` (https://github.com/plone/plone.app.upgrade/pull/251).

We could also remove a `CMFFormController` line from `plone.app.testing`, but it's not in the way.
Could be done once we make a `plone.app.testing` branch for Plone 6 only.

In `plone.app.upgrade` I remove the `portal_form_controller` tool.
We could also conditionally remove it: check if the tool is an instance of `ZODB.broken.Broken`.
This is to support the theoretical case that someone explicitly adds the `CMFFormController` package, because it is still needed in a specific site.
But same could be done for `portal_quickinstaller`, and we don't do it there either.

Once green and approved, I can do the merges. The two PRs should be merged, and then on coredev we can keep using the master branches, and only merge the part of this PR that removes `CMFFormController` from the `test-eggs` and sources/versions.


jensens:
Removal is fine IMO, if this is an issue we can refactor later. But I do not expect it to be one.

I would say, just merge if ready, LGTM (also the connected PRs)

jensens (approved):

mauritsvanrees:
I have merged all.

PR #698, opened 2021-01-26
use latest setuptools/pip/buildout

jensens:
Works for me, lets see what GHActions and Jenkins thinks.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,10 +1,11 @@
 setuptools==54.0.0
-zc.buildout==2.13.3
-wheel
+zc.buildout==3.0.0b1
+pip==21.0.1
 
 # Windows specific down here (has to be installed here, fails in buildout)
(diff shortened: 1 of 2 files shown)


requirements.txt
```diff
@@ -1,7 +1,7 @@
-setuptools==50.3.0
-zc.buildout==2.13.3
-wheel
+setuptools==52.0.0
+zc.buildout==3.0.a2
```
tisto: I guess this should be 3.0.0a2. Do we really want to depend on a zc.buildout alpha? That should be fine for the 6.0 branch though...

jensens: Looks like it works despite the typo.

This is primary a quick test to see if it works. And it targets the 6.0 branch anyway.

jensens:
Yeah! @gotcha lots of hugs!

PR #699, opened 2021-02-02
Uses TZ = UTC in buildout

wesleybl:
This ensures that tests using datetime will work locally on machines using TZ = GMT

Branch 6.0


jensens:
This one broke CI on Python 3.8 (only) https://jenkins.plone.org/job/plone-6.0-python-3.8/185/
@wesleybl can look at it and fix? Otherwise I'll revert.

wesleybl:
I think it's best to revert for now. When I have time I take a look at it.

wesleybl:
By the way, shouldn't PRs in branch 5.2 have testing with Python 3.8? Or is it still not supported?

wesleybl:
I see that: 

https://github.com/plone/Products.CMFPlone/blob/5.2.x/setup.py#L28

PR #693, opened 2020-12-30
Uses TZ = UTC in buildout

wesleybl:
This ensures that tests using datetime will work locally on machines using TZ = GMT.

PR #701, opened 2021-02-05
Set timezone in bin/test.

mauritsvanrees:
Calling time.tzset is needed on Python 3.8,
otherwise the TZ environment variable that we add has no effect.
See discussion starting with this comment:
https://github.com/plone/buildout.coredev/pull/693#issuecomment-774160597


mauritsvanrees:

And I will start the 3.8 job manually.

PR #707, opened 2021-02-17
generate plone-compile-resources, it's in plone.staticresources now

fredvd:


mauritsvanrees:
This should have no effect: `plone.staticresources` is already in the `${test:eggs}`. Locally at my machine the script is created.

Can others check if they have the script?

thet:
@mauritsvanrees @fredvd i also have the script generated.
@fredvd isn't it generated for you? if so, let's find the real issue.

jensens:
I am with @thet and do not think it is needed.

fredvd:
I agree it should not be needed as well, but there is an edge case where it does not seem to be generated, I ran into it, somebody else did (reported on community) and I have no clue yet why.  :-S

PR #710, opened 2021-02-18
5.2: move constraints generation code into release dir with tox

mauritsvanrees:
- Move `create-constraints.py` out of the buildout config and into a new `release` directory. I am guessing here that no one actively uses the locally created `constraints.txt`. I could be wrong.
- Add `release/tox.ini`. You can call it with `tox -c release/tox.ini -p auto`. It takes care of a few things:
- Use a small Python 2 buildout to parse a `release/constraints2.txt`.
- Use a small Python 3 buildout to parse a `release/constraints3.txt`.
- Use a new script `combine-constraints.txt` to combine these two into `release/constraints.txt` fit for both Python versions, by adding modifiers like `python_version < "3.0"` where needed.

And then it does a few things that help me as release manager:

- Use `pip2 download` to download *all* packages in `constraints.txt` and put them in a `release/dist` directory. This only downloads Python 2 compatible package versions. Useful is that this complains and quits when a package is pinned to a version that is incompatible with Python 2. I updated a few version pins in `versions.cfg` for this.
- Use `pip3 download` to download all Python 3 compatible packages from `constraints.txt`.
- Copy `versions.cfg` and a few other files into the `release/dist` directory so I can `scp` this to `dist.plone.org`.

We could do this for Plone 6 as well, but there it is far less tricky, because everything is Python 3.


jensens:
Like! And yes, for Plone 6 we need it too!

PR #712, opened 2021-02-20
Pin cffi and cryptography to avoid issues with M1 macs and missing ru…

tisto:
…st complilers

### DIFF
--- a/versions.cfg
+++ b/versions.cfg
@@ -60,6 +60,14 @@ selenium = 3.141.0
 i18ndude = 5.3.4
 msgpack = 0.6.2
 
+# plone.restapi
+
+# cryptography 3.4 requires a rust compiler installed on the system:
+# https://github.com/pyca/cryptography/blob/master/CHANGELOG.rst#34---2021-02-07
+cryptography = 3.3.2
+
+# cffi 1.14.3 fails to install on apple m1
+cffi = 1.14.4
 
 ##############################################################################
 # Zope world dependencies


tisto:

jensens:
I propose to put those in conditional sections, because on other platforms I prefer the more secure rust based installation. 
http://www.buildout.org/en/latest/topics/variables-extending-and-substitutions.html#conditional-configuration-sections

mauritsvanrees:
Actually, `cryptography` is pinned to a much earlier 2.9.2. But this is only in `tests.cfg`. Earlier I already thought about just moving those to `versions.cfg`. It might be time for that. I will see about creating a PR for that.
The `constraints.txt` that I compile already has those versions included.

Meanwhile, can you try `cryptography` 3.4.6? I see this [includes a wheel for arm](https://pypi.org/project/cryptography/3.4.6/#files).

mauritsvanrees:
Superseded by https://github.com/plone/buildout.coredev/pull/713

tisto:
@mauritsvanrees I can try cryptography later today on my m1 when I am at home again. Though, cryptography fails because newer versions require a rust compiler installed on the system. This has nothing to do with the M1 or Python 2. We ran into this problem on our Ubuntu machines with Python 3 as well.

In order to not break existing buildout installations on prod servers, we should pin cryptography to a non-rust version IMHO. For Plone 6 we can make the rust compiler a system requirement and then go with a newer cryptography version.

jensens:
@tisto, if there is a binary wheel no compiler is needed. In my perfect world all packages are available as bin-wheels for all major platforms, so no compiler at all is needed to install Plone.

tisto:
@mauritsvanrees @jensens cryptography 3.4.6 works on my M1. Though, it fails on our Ubuntu Jenkins nodes because the rust compiler is not installed: 

`ModuleNotFoundError: No module named 'setuptools_rust'`

No clue why it compiles and does not take the bin-wheel. I did not investigate this further though.

As said before, I think it would be better for most users to downgrade cryptography to the last non-rust version. This would give us some time to adapt. Just my 2c...

mauritsvanrees:
@tisto Thanks for trying. In 3.4.6 I see most [wheels](https://pypi.org/project/cryptography/3.4.6/#files) have `cp36` in their name, so that probably only supports CPython 3.6. There are a few others.
So yes, we may run into problems with this.
I will pin 3.3.2.

PR #715, opened 2021-03-03
Use zope.publisher and zope.schema from Zope 4.5.5.

mauritsvanrees:
I did not trust these versions enough to include them in 5.2.4, fearing some breakage.
But let's get them in now.


mauritsvanrees:

PR #721, opened 2021-05-14
Update calmjs.parse to fix a deprecation warning

ale-rt:
```
/home/ale/.buildout/eggs/calmjs.parse-1.2.4-py3.8.egg/calmjs/parse/io.py:7: DeprecationWarning: Using or importing the ABCs from 'collections' instead of from 'collections.abc' is deprecated since Python 3.3, and in 3.9 it will stop working
  from collections import Iterable    
```

### DIFF
--- a/versions.cfg
+++ b/versions.cfg
@@ -95,7 +95,7 @@ zope.schema = 6.1.0
 
 ##############################################################################
 # External dependencies
-calmjs.parse = 1.2.4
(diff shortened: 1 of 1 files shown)


ale-rt:
See also https://github.com/calmjs/calmjs.parse/commit/00a2d6f81ed7196eff201459a71ba7c2a49122a8

PR #722, opened 2021-05-26
Uses ignore-develop = True in omelette

wesleybl:
Development packages are already in the src. Including them in the omelette will leave them in two places.

### DIFF
--- a/core.cfg
+++ b/core.cfg
@@ -135,7 +135,7 @@ eggs =
 
 [packages]
 recipe = collective.recipe.omelette
-ignore-develop = False
+ignore-develop = True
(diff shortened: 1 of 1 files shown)


wesleybl:

mauritsvanrees (changes requested):
I am very much against this. I often grep in `parts/packages` and expect this to contain all code.

With this merged, I would need to grep in both `parts/packages` **and** `src` and get duplicate results because `src` will contain old checkouts of packages for which a release is in `parts/packages`.

wesleybl:
Good point @mauritsvanrees.

PR #723, opened 2021-05-26
newer version of a bunch of packages

jensens:


jensens:

mauritsvanrees:
This PR is two months old. Probably better to try again after we have [Zope 5.3 in our versions](https://github.com/plone/buildout.coredev/pull/730).

Also, we can look at the [actions artifact](https://github.com/plone/buildout.coredev/actions/runs/1116277824) to see which versions may be updated.

PR #730, opened 2021-08-04
Plone 6: use Zope 53

mauritsvanrees:


mauritsvanrees:

mauritsvanrees:
Almost okay, but there are failures in GenericSetup. See https://github.com/zopefoundation/Products.GenericSetup/issues/109

jensens:
can we use an older version of GS for now and keep using the other releases?

mauritsvanrees:
> can we use an older version of GS for now and keep using the other releases?

No: there is no older GS version for which the tests pass on Zope 5.3.

One place in Plone where it could go wrong, is in dexterity fti behaviors. Since `plone.dexterity` 2.6.1, available in Plone 5.1.4, for new FTIs in [Python 3 the behaviors property is a `ulines`](https://github.com/plone/plone.dexterity/commit/1e0699034b04170b4dca3e5cc11a3a45cb9c00e2) so on Py 3 it will contain strings in every Zope version. But on Py 2 it is still `lines`. So the following might cause problems (theoretically anyway, I have not tried it):

- Create a Plone 5.2 Python 2 site (or migrate a 5.1 site). The Document FTI has a behaviors property of type `lines`, so this contains bytes.
- Import a `types/Document.xml` which adds behavior `foo`.
- Migrate the database to Python 3. Behaviors is still a lines property with bytes.
- Migrate the code to Plone 6.0 and Zope 5.3. Behaviors is still a lines property with bytes.
- Import the same `types/Document.xml` which adds behavior `foo`.
- Behaviors is still a lines property, but it will have a mix of bytes and strings, and `foo` will be in there twice: once as bytes, once as string.

I think my [GenericSetup PR 110](https://github.com/zopefoundation/Products.GenericSetup/pull/110) would be fine.

mauritsvanrees:

PR #731, opened 2021-08-05
Pin zope.schema back to 6.0.0.

mauritsvanrees:
6.1.0 causes problems with booleans.
For example, easyform cannot mail anymore.
See https://community.plone.org/t/plone-5-2-5-soft-released/14150/9?u=mauritsvanrees


mauritsvanrees:

PR #733, opened 2021-08-10
Copy and adapt the release directory from coredev 5.2.

mauritsvanrees:
This contains a tox.ini and create-constraints script which helps in preparing a (pending) directory on dist.plone.org.
This replaces `scripts/create-constraints.py` and the `constraints` section from `core.cfg`.

This should be fine to merge, and should only be interesting for the release managers. But maybe someone is using the old `constraints` section. So here is your chance to speak up. :-)
I already did the same on 5.2 months earlier though.

I won't start tests on Jenkins, because this will have no influence. But the GH Actions should still pass.


release/create-constraints.py
```diff
@@ -0,0 +1,46 @@
...
+from zc.buildout import buildout
+
+import os
+import sys
+
+
+# List packages that we should not include.
+# Originally this was for packages that are pinned by Zope
+# but cannot be pip installed on Py 2 or on Py 3.
+BLACKLIST = [
```
jensens: Well, better `DENYLIST`?

jensens:
```suggestion
DENYLIST = [
```

release/create-constraints.py
```diff
@@ -0,0 +1,46 @@
...
+# Get the constraints from the version pins.
+# Nice: the versions get set directly on the config.
+# Note: this works like a dictionary, but is a class 'zc.buildout.buildout.Options'.
+versions = config.versions
+
+with open(constraints_file, "w") as cfile:
+    cfile.write("# File created by {}\n".format(__file__))
+    cfile.write("# Constraints parsed from {}\n".format(config_file))
+    for package, version in sorted(versions.items()):
+        if package in BLACKLIST:
```
jensens:
```suggestion
        if package in DENYLIST:
```

jensens (approved):
except for the tiny naming issue, LGTM.

release/create-constraints.py
mauritsvanrees: Ah, yes, same as in the hotfix. I will change it and then merge.

PR #739, opened 2021-09-22
Create plone-register-icons and plone-register-flags scripts.

thet:
Add scripts to register new bootstrap-icon and svg-country-flags icons in plone.staticresource.

Note: This does not happen very often, only if a new release - mainly from bootstrap-icons - with new icons is released.


thet:
I do need this scripts right now. Merging.

PR #753, opened 2021-11-20
Use latest pip, buildout, setuptools.

mauritsvanrees:
zc.buildout had problems with pip greater than 21.0.1, but that was solved in recent 3.0.0b3.

I do get lots of warnings about invalid versions, but they seem to be from versions that we don't use:

    .../lib/python3.9/site-packages/pkg_resources/__init__.py:116:
    PkgResourcesDeprecationWarning:
    1.2.7-20070827-preview is an invalid version and will not be supported in a future release

That is elementtree, which we don't use.
It also complains about versions where the package name is included: `plone-api-1.4`.

Aha, buildout is going through the cached downloads.
I have this in my `~/.buildout/default.cfg`:

```
[buildout]
index = https://pypi.org/simple/
eggs-directory = /Users/maurits/shared-eggs
download-cache = /Users/maurits/cached-downloads
abi-tag-eggs = true
```

Downloaded packages are saved in `cached-downloads/dist`, which has about 5600 packages.
So Buildout finds a lot of old packages that have versions that cannot be strictly parsed,
so they get a legacy version.  Maybe it was all fine at the time, but current setuptools complains.

It is just a warning, so we could ignore it.
But for me it takes about ten seconds before Buildout is finished going through this directory.
And a bit later it does it again.

Solution for me: rename the `dist` directory to `olddist` so I still have it in case packages disappear,
and create an empty `dist` directory.
Now buildout takes 36 seconds instead of 56.


mauritsvanrees:

mauritsvanrees:
Rebased and force pushed.  Let's first see what the GitHub checks say.

mauritsvanrees:
I have upgraded FormEncode, which gave a problem on Mac. The Github Actions now pass.

In https://github.com/plone/diazo/pull/83 I am removing FormEncode, which is only a test dependency of Diazo. But the new version in the current PR should be fine.

Meanwhile:

jensens (approved):

ale-rt (approved):

PR #754, opened 2021-11-20
Use plone.restapi 7 on Py 2, plone.restapi 8 on Py 3.

mauritsvanrees:
[I accidentally committed and pushed this directory on the coredev 5.2 branch. I immediately reverted.]

This means two different branches in sources.cfg.
That will be confusing when testing a PR... So this might not be a good idea after all.

See also discussion in https://github.com/plone/plone.volto/issues/22


mauritsvanrees:
Let's see what happens anyway.

mauritsvanrees:
Let's close this, it is not going to happen.

PR #755, opened 2021-11-28
Add git repo and version for plone.autoinclude

ale-rt:
Preparatory work for #3339.


ale-rt:
This can be safely merged even now, it is just a version pin and some info about plone.autoinclude git repo.

jensens:
It really does not hurt!

PR #762, opened 2021-12-29
build(versions): Fix local devel buildout error

rpatterson:
$ bin/buildout
    ...
    While:
      Installing.
      Uninstalling plonesite-volto.
      Installing recipe collective.recipe.plonesite.
      Getting distribution for 'collective.recipe.plonesite'.
    Error: Picked: collective.recipe.plonesite = 1.12.0

    The `collective.recipe.plonesite` egg does not have a version pin and `allow-picked-versions = false`.

    To resolve this, add

        collective.recipe.plonesite = 1.12.0

    to the [versions] section,

    OR set `allow-picked-versions = true`.

### DIFF
--- a/versions.cfg
+++ b/versions.cfg
@@ -28,6 +28,7 @@ cachecontrol = 0.12.6
 click = 8.0.1
 click-default-group = 1.2.2
 collective.recipe.omelette = 1.1.0
+collective.recipe.plonesite = 1.12.0
 collective.recipe.vscode = 0.1.7
 Genshi = 0.7.5
 incremental = 21.3.0


ale-rt (commented):
I am -0 merging this.
The package `collective.recipe.plonesite` is not mentioned anywhere else in the coredev buildout.
I would add this pin to your local buildout.

mauritsvanrees:
I agree with @ale-rt.
In some cases it makes sense to add a pin for a package that is not actually used by the coredev buildout, but I would rather remove some of those.

rpatterson:
My apologies, I think I got myself mixed up between `buildout.coredev` and the `volto/api/` buildout and convinced myself that `collective.recipe.plonesite` *was* used in the former.  I see now that it isn't at all.

PR #765, opened 2022-01-13
Update Pillow to 9.0.0

jensens:
Lets see what breaks, should give a nice performance boost if libjpeg-turbo is available in the target system.


jensens:

ale-rt (approved):

PR #767, opened 2022-01-19
Check release/6.0-dev branch

mauritsvanrees:
Not for merging, only for testing.

We are getting close to a third alpha, so let's check the branch that I actually make a release from.
Note that this has Zope 5.4 release instead of a checkout. This is the prime reason that I want an extra test.


mauritsvanrees:

mauritsvanrees:
Only one `plone.rest` failure, which would be fixed if we check it out or have a new release. Good!
Closing.

PR #775, opened 2022-02-17
5.2: always checkout Zope 4.x.

mauritsvanrees:
And extend the versions from this branch.
Then we can know of any incompatibilities before a new Zope release is created.
We have done this in Plone 6 coredev for a few months now.

### DIFF
--- a/versions.cfg
+++ b/versions.cfg
@@ -2,7 +2,10 @@
 # This file is part of buildout.coredev repository: https://github.com/plone/buildout.coredev
 # The plone release team is responsible for it,
 # if you have suggestions, please open an issue at: https://github.com/plone/buildout.coredev/issues
-extends = https://zopefoundation.github.io/Zope/releases/4.7/versions.cfg
+# Based on latest development Zope:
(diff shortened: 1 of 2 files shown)


ale-rt (approved):

mauritsvanrees:

versions.cfg
```diff
@@ -83,14 +86,9 @@ zope.keyreference = 4.2.0
 zope.mkzeoinstance = 4.1
 zope.password = 4.3.1
 
-# Newer versions than from the current Zope 4.7.  Remove after Zope updates it.
```
wesleybl, line 86: Was this information not helpful? If a version of Plone is released before a version of Zope, how will you know which packages to pin?

mauritsvanrees: I will compare the Zope versions from 4.7 with 4.x and add those to our own versions.
I will likely do this only on the [5.2-dev](https://github.com/plone/buildout.coredev/blob/release/5.2-dev/versions.cfg) branch from which I make the releases.

PR #779, opened 2022-02-24
Plone 6: Update several versions to latest.

mauritsvanrees:
With help from `bin/versioncheck`.
We may want to actually remove some.

Removed `Products.contentmigration` because it is Python 2 only.


mauritsvanrees:

ale-rt (approved):

PR #780, opened 2022-03-11
Pin plone.autoinclude 1.0.0a4

ale-rt:
I am not sure if it makes sense to have this in the 5.2 branch.


mauritsvanrees (commented):
I would not add this.

PR #783, opened 2022-03-23
Add webresource to versions.cfg.

thet:
It is a dependency of CMFPlone 6, PLIP 3211


rnixx:
Let me drop another release of webresource first. It contains then the GracefulRenderer which is required to work with addons defining broken resources.

mauritsvanrees (approved):

PR #786, opened 2022-04-07
Use fresh zc.buildout 3.0.0rc3 and setuptools 62.0.0.

mauritsvanrees:
This should work now. See https://github.com/buildout/buildout/issues/600#issuecomment-1092113506


mauritsvanrees:

PR #787, opened 2022-04-13
Updated package versions, removed several unused ones. [6.0]

mauritsvanrees:
I started doing this before https://github.com/zopefoundation/Zope/pull/1028 was merged, but was too late.

Since that merge, I copied over the removed Zope versions directly in the 6.0 branch, to fix the build. We can remove more version pins from there.

But here is some initial cleanup. Note: also includes a `setuptools` increase, so you may want to run `bin/pip install --force-reinstall -r requirements.txt`

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,4 +1,4 @@
-setuptools==62.0.0
+setuptools==62.1.0
 zc.buildout==3.0.0rc3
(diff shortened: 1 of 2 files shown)


mauritsvanrees:

PR #788, opened 2022-04-13
Maurits big cleanup

mauritsvanrees:
Earlier today, Zope removed lots of version pins for packages that core Zope should not care about. We should do the same for Plone. This PR does the following:

- Cleanup of old no longer used settings and files.
- Completely restructure `versions.cfg`, splitting non-core versions out into `versions-extra.cfg`
- To keep a bit of structure in the still long list in `versions.cfg`, I split them into four sections based on GitHub organisations: plone, collective, zope/zodb/z3c/repoze, and other.
- I added `bare.cfg` plus an extra GHA yaml file to check the the pins in `versions.cfg` are enough to create a Plone instance plus a `bin/test` script.


jensens:
Overall big love for this! But I would not mix up core + ecosystem packages (like mosaic). I tend to add a versions-ecosystem file unless those are not officially core. What do you think?

mauritsvanrees:
> Overall big love for this! But I would not mix up core + ecosystem packages (like mosaic). I tend to add a versions-ecosystem file unless those are not officially core. What do you think?

That can work.
But in this PR, there are not yet any ecosystem packages. All versions in `versions.cfg` are used while creating `bin/instance` and `bin/test`. For example I did not expect `plone.tiles` in here, but `plone.restapi` pulls it in as test dependency.

mauritsvanrees:
I have added `versions-ecosystem.cfg` and updated  `ecosystem.cfg` to use this.

mauritsvanrees:
I have improved the GitHub Actions. Should cache better the next time.
Now Jenkins:

ale-rt (approved):
Wonderful job :)

Some consideration you might ignore or that can be handles another time:
- should core.cfg extend bare.cfg?
- should we also have a sources-extra.cfg?

mauritsvanrees:
> * should core.cfg extend bare.cfg?

That is an option. Okay, I have changed it.

> * should we also have a sources-extra.cfg?

I think it is clearer to have all sources in one file, so developers do not have to hunt for where a source is defined. I don't expect that we will define many extra sources.
I have reordered `sources.cfg` a bit now, putting the non-eggs apart, and removing sources that are already define in the Zope sources.

mauritsvanrees:
Meanwhile trying to improve the caching of GHA further. Previous run was all over the place for how long a run took, some around 2 minutes, some 15, some almost 30:

Screenshot 2022-04-14 at 10 31 24

I am trying this now in all yaml files:

```
- name: Cache eggs
  uses: actions/cache@v3
  with:
    path: |
      eggs
    key: eggs-${{ matrix.python-version }}-${{ runner.os }}-${{ hashFiles('versions*.cfg') }}
    restore-keys: |
      eggs-${{ matrix.python-version }}-${{ runner.os }}-
      eggs-${{ matrix.python-version }}-
```

mauritsvanrees:
Another run after my cache changes shows around 15 minutes for each, as none of them now find a cache:

Screenshot 2022-04-14 at 10 51 27

I did another small change, configuring the vscode recipe, and in the next run all should be a lot faster.

mauritsvanrees:
That is better. Longest was 6 minutes:

Screenshot 2022-04-14 at 11 02 02

Cache extraction works:

```
Cache restored from key:
eggs-3.7-Linux-4c6074f594712aef4a7d92010ae65d7d4dd84e7c7c7f268628e46071eaa03b65
```

ale-rt (approved):

ale-rt:
Merge this at will :)

mauritsvanrees:
Merged. Thanks for the reviews.

sources.cfg
```diff
@@ -164,26 +155,30 @@ Products.ExternalEditor             = git ${remotes:zope}/Products.ExternalEdito
...
-Products.Sessions                   = git ${remotes:zope}/Products.Sessions.git pushurl=${remotes:zope_push}/Products.Sessions.git branch=master
-Products.SiteErrorLog               = git ${remotes:zope}/Products.SiteErrorLog.git pushurl=${remotes:zope_push}/Products.SiteErrorLog.git branch=master
 Products.statusmessages             = git ${remotes:plone}/Products.statusmessages.git pushurl=${remotes:plone_push}/Products.statusmessages.git branch=master
 Products.validation                 = git ${remotes:plone}/Products.validation.git pushurl=${remotes:plone_push}/Products.validation.git branch=master
-Products.ZCatalog                   = git ${remotes:zope}/Products.ZCatalog.git pushurl=${remotes:zope_push}/Products.ZCatalog.git branch=master
 Products.ZopeVersionControl         = git ${remotes:zope}/Products.ZopeVersionControl.git pushurl=${remotes:zope_push}/Products.ZopeVersionControl.git branch=master
+repoze.xmliter                      = git https://github.com/repoze/repoze.xmliter.git pushurl=git@github.com:repoze/repoze.xmliter.git branch=master
```
gforcada, line 168: We could use here the `${remotes:zope}` :)

versions-ecosystem.cfg
```diff
@@ -0,0 +1,22 @@
+# These are extra version pins for packages that are not part of core or tests.
+# But they are often used modules and we want to have canonical versions of them.
+# Note: this is ony lightly tested, and we cannot give guarantees that these versions work without error.
+# Read the documentation of the individual packages, or ask the maintainers.
+[versions]
+# collective.z3cform.datagridfield = 2.0.2
+# collective.z3cform.datetimewidget = 1.2.9
```
gforcada, line 7: you kept these pins commented as they are not needed, but might be worth keeping them around as we might want to uncomment them eventually? 🤔

gforcada:
Sorry to be late on the party 😅 I was on vacations with my kids 🎉 

I could only mentions these two minor issues/questions/fixes, great job @mauritsvanrees ! 💯

sources.cfg
mauritsvanrees: That would then need to be `remotes:repoze`, which does not exist yet. And since it is the only package we use from repoze, let's not add the extra indirection.

versions-ecosystem.cfg
mauritsvanrees: I kept them because it would be good to have version pins for these, as they were in the 5.2 pins as well. But there are no released versions of these two packages yet that work on current Plone 6.
Some work is done here:

- https://github.com/collective/collective.z3cform.datetimewidget/pull/12
- https://github.com/collective/collective.z3cform.datagridfield/pull/120

sources.cfg
gforcada: oh my 🤦🏾 sorry, I was not reading properly 😓 sorry for the noise!

PR #791, opened 2022-04-19
Products.ZCatalog=6.2

### LINKED ISSUE #3432: Remove the treatment of the "created" parameter in the search view
To fix #3007, @frapell did PR #3012. However, the problem occurs with any parameter that uses record notation in the URL. So I made PRs https://github.com/zopefoundation/Products.ZCatalog/pull/129 and https://github.com/zopefoundation/Products.ZCatalog/pull/130 in `ZCatalog`, to handle any record parameter.

Therefore, we must remove the treatment of the `created` parameter:

https://github.com/plone/Products.CMFPlone/blob/954942a1ec95d9c9472ecf791d8d55ca096b33bb/Products/CMFPlone/browser/search.py#L121-L124

when Plone 6/Plone 5.2 uses a version of Zope that contains these PRs.

Probably when Plone 6 is using Zope > 5.4 and Plone 5.2 is using Zope > 4.7

wesleybl:
`Products.ZCatalog` 6.2 is not pinned in current Zope version 5.5.1. First step to fix plone/Products.CMFPlone#3432 in Plone 6

### DIFF
--- a/versions.cfg
+++ b/versions.cfg
@@ -156,6 +156,7 @@ Products.PythonScripts = 4.13
 Products.Sessions = 4.12
 Products.SiteErrorLog = 5.5
 Products.StandardCacheManagers = 4.1.1
+Products.ZCatalog = 6.2
 Products.ZopeVersionControl = 2.0.0
 repoze.xmliter = 0.6.1
 z3c.caching = 2.2
@@ -227,3 +228,5 @@ robotframework =
     Since 3.2.x the robot.parsing module API has breaking changes, thus we need to adopt before upgrading.
 Pillow =
     9.1.0 gives a test failure in plone.scale due to an extra DeprecationWarning.  See https://github.com/plone/plone.scale/issues/49
+Products.ZCatalog =
+    6.2 is not pinned in current Zope version 5.5.1. First step to fix https://github.com/plone/Products.CMFPlone/issues/3432


wesleybl:

jensens:
I wonder why we need a comment here. Otherwise +1

wesleybl:
@jensens I tried to follow the way that @mauritsvanrees  does. I think the comment can help to know when to remove the pin as Zope should take care of it.

jensens:
I know, I introduced the comment feature for `plone.versioncheck`. So, if Zcatalog ever goes into Zope core we will notice.

wesleybl:
> I know, I introduced the comment feature for plone.versioncheck. So, if Zcatalog ever goes into Zope core we will notice.

@jensens @mauritsvanrees  so, when you are going to generate a `Plone` release, are versions indicated that are the same as the `Plone` release and the `Zope` release, so they can be removed from the `Plone` `versions.cfg`? In these situations, then, are comments like this really unnecessary?

Can we remove this comment? https://github.com/plone/buildout.coredev/blob/9506e23f64024f55cdd9c1b0b03066273fdc1d62/versions.cfg#L91

mauritsvanrees:
This is mostly relevant not when I make a new Plone release, but when there is a new Zope release. Then I can look in the short list of older/newer packages, and use the extra information in the annotations to determine if I can remove one of these pins.

So please keep the comment: it encourages people (including myself) to add a version annotation at the moment when they add the version pin, instead of possibly months later when I look at the list and have no clue and have to do detective work to figure out if it is safe to remove a version pin.

wesleybl:
> So please keep the comment: it encourages people (including myself) to add a version annotation at the moment when they add the version pin

@mauritsvanrees the comment I pointed out is from `Plone` 5.2. In `Plone` 6.0 it has been removed. Should we add it in `Plone` 6?

mauritsvanrees:
Yes please.

PR #792, opened 2022-05-02
Upgrade to robotframework 3.2, using robotsuite 2.3.

mauritsvanrees:
We may be able to do a much larger version upgrade, see `plips/plip-robotframework.cfg`, but this is an important first step.

See https://github.com/collective/robotsuite/pull/23 for the changes that Asko merged today.


mauritsvanrees:

mauritsvanrees:
As a bonus, this brings us a good step closer to supporting Python 3.10, which is one reason why I pushed this. (Previous `robotframework` had an ImportError.)
Locally with Python 3.10 and this branch, I can create a Plone Classic site, and run `bin/test -u` and `bin/test --all -s Products.CMFPlone` without failures.
Undoubtedly, there will be a few packages that have ImportErrors or other incompatibilities, but this already sounds very good!

gforcada (approved):

PR #797, opened 2022-06-07
Use latest selenium and robotframework.

mauritsvanrees:
Selenium Python changes are here:
https://github.com/SeleniumHQ/selenium/blob/trunk/py/CHANGES
Notable, already in 4.1.4:

* Add support Chrome 100 and remove for Chrome 97
* Adding Chrome DevTools 101 and removing 98


mauritsvanrees:

mauritsvanrees:
Does not seem better or worse than the current versions.

fredvd (approved):
LGTM tested those changes locally yesterday. Saw no direct benefits either, but there are quite some bugs fixed, lets keep up to date.

PR #805, opened 2022-07-08
use branch 1.x of plone.app.querystring

erral:
See https://github.com/plone/plone.app.querystring/pull/112


petschki:
I've updated it directly on `5.2` ... sorry, saw your PR too late

PR #807, opened 2022-07-12
Checkout more Zope Products on 6.0.

mauritsvanrees:
Plus three small version updates.

They are already in the checkouts of 5.2. I want to have releases of them, but we should check them on 6.0 first.


mauritsvanrees:

PR #816, opened 2022-09-30
Downgrade pip, upgrade setuptools.

mauritsvanrees:
With `pip>=22.2` that we have been using, buildout fails to check the python-requires of a package, so you may get incompatible versions. See https://github.com/buildout/buildout/issues/613 A fix is underway, but that can take a while.
So downgrade pip to 21.
Do try the latest setuptools.

The error you get is:

```
$ bin/buildout
Requires-Python support missing.

  File "/Users/maurits/community/plone-coredev/6.0/lib/python3.10/site-packages/zc/buildout/patches.py", line 52, in patch_PackageIndex
    from pip._internal.index.collector import HTMLPage
ImportError: cannot import name 'HTMLPage' from 'pip._internal.index.collector' (/Users/maurits/shared-eggs/cp310/pip-22.2-py3.10.egg/pip/_internal/index/collector.py)
```

Buildout *does* continue after that, but as said: it is missing the python-requires functionality.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,7 +1,11 @@
-setuptools==63.2.0
-zc.buildout==3.0.0rc3
-pip==22.2
+# With pip>=22.2, buildout fails to check the python-requires of a package,
+# so you may get incompatible versions.
+# See https://github.com/buildout/buildout/issues/613
+# So we pin an older one.
+pip==22.1.2
+setuptools==65.4.1
 wheel==0.37.1
+zc.buildout==3.0.0rc3
 
 # Windows specific down here (has to be installed here, fails in buildout)
 # Dependency of zope.sendmail:
--- a/versions.cfg
+++ b/versions.cfg
@@ -13,10 +13,10 @@ extends = https://zopefoundation.github.io/Zope/releases/5.6/versions.cfg
 [versions]
 # Basics
 # !! keep in sync with requirements.txt !!
-setuptools = 63.2.0
-zc.buildout = 3.0.0rc3
-pip = 22.2
+pip = 22.1.2
+setuptools = 65.4.1
 wheel = 0.37.1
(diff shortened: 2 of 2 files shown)


mauritsvanrees:

jensens:
Please do not downgrade for pip based installs in requirements! Newer pip is way faster and more robust.

mauritsvanrees:
I have not yet noticed that pip 22 is faster. Do you know why it would be faster?
Looking through the [changelog](https://pip.pypa.io/en/stable/news/).

Maybe this one? "Significantly speed up isolated environment creation, by using the same sources for pip instead of creating a standalone installation for each environment. 11257"? But that sounds like it would only be for initially creating the environment, which should not amount to "way faster" overall. But I might be misinterpreting.

mauritsvanrees:
I try this:

```
$ time bin/pip install Products.CMFPlone -c https://dist.plone.org/release/6.0.0b2/constraints.txt
```

I do this for both 22.1 and 22.2, in separate environments. Then I throw the envs away and recreate them. So the caches have everything that is needed. Then try again:

- With pip 22.2 it takes 52 seconds.
- With pip 22.1 it takes 54 seconds.
- A second try, with everything already installed in the env, takes 7 seconds in both cases.

So in this single test on my machine (Mac, Monterey since this week) the difference is negligible.

mauritsvanrees:
I am inclined to merge this, otherwise buildout can install package versions that are not compatible with your Python version. Compared to a possibly slower experience when you only use pip, I would say a merge is preferable.

This [PR](https://github.com/buildout/buildout/pull/616) should fix the buildout problem, but some test failures remain, likely unrelated.

jensens:
+1 for merge to fix the immediate problem and then ASAP fix buildout and undo this PR

PR #833, opened 2022-12-13
Checkout Zope, extend Zope master versions.

mauritsvanrees:
The Zope code has no changes yet.
We usually have it checked out, but I had switched this off a while before the final release, so we really tested with a Zope release.


mauritsvanrees:

gforcada (approved):

PR #839, opened 2023-01-16
Use latest setuptools, zc.buildout, wheel, on Python 3 Plone 5.2.

mauritsvanrees:
This means we install different versions for these packages on Python 2 and 3.  I don't like this. But some packages are getting harder to install on Python 3 with older versions of these base packages. See my adventures on installing zope.container when I switched from Python 3.8.13 to 3.8.16 on Mac: https://github.com/zopefoundation/zope.container/issues/48 So I think we are better off with the latest versions.

Technically we do not use latest setuptools, which would be 66.0.0. This is very new, and I do not trust the breaking change of no longer supporting versions that do not conform to PEP 440: https://github.com/pypa/setuptools/blob/v66.0.0/CHANGES.rst

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,6 +1,10 @@
-setuptools==42.0.2
-zc.buildout==2.13.8
-wheel==0.37.1
+setuptools==42.0.2; python_version < '3.0'
+zc.buildout==2.13.8; python_version < '3.0'
+wheel==0.37.1; python_version < '3.0'
+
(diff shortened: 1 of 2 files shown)


mauritsvanrees:

ale-rt (approved):
Sure thing!

gforcada (approved):
LGTM 👍🏾

PR #841, opened 2023-02-01
Update zope branches in sources.cfg. [5.2]

mauritsvanrees:
A lot of master branches are being updated to drop Python 2.7 support, so we want maintenance branches. I removed several source definitions that were duplicate from the one in Zope/sources.cfg, or that are really not used anymore, like five.pt and ZODB3. I am updating the Zope 4.x sources.cfg as well, and don't want to do this in two places.
Related PR, which would best be merged first: https://github.com/zopefoundation/Zope/pull/1099

I will make the coredev PR a draft PR for now.

### DIFF
--- a/sources.cfg
+++ b/sources.cfg
@@ -1,8 +1,8 @@
 [buildout]
 extends =
     https://raw.githubusercontent.com/zopefoundation/Zope/4.x/sources.cfg
-
-# auto-checkout has moved to checkouts.cfg so it's easier to extend sources.cfg
+# When Zope is in auto-checkout, and you are editing the sources, you want this instead:
+#    src/Zope/sources.cfg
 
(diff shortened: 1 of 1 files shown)


gforcada:
Sounds like a good idea to rely on zope for their sources configuration 👍🏾 

Doesn't that go at odds with the idea of having `requirements.txt` leading (i.e. `pip` based workflow) rather than `versions.cfg` ? 🤔 

Or is it irrelevant? 😅

mauritsvanrees:
This PR is for Plone 5.2. I am not planning to make requirements.txt the canonical place there.

I have merged my Zope PR, so we can test now on Jenkins, although it should not have an influence as long as we do not checkout any further packages.

PR #843, opened 2023-02-10
Checkout PortalTransform 3.x

ale-rt:
We want to have a 4.x release that depends imports from Products.CMFPlone (which is not listed in the requirements) with imports from plone.base (that will be added to the requirements).


ale-rt:
Required for https://github.com/plone/Products.PortalTransforms/pull/52

mauritsvanrees (approved):

PR #846, opened 2023-03-08
Undo checkout of plone.app.linkintegrity.

mauritsvanrees:
The Plone 6.0 jobs are failing since yesterday.
For Py 3.11, this is the first failing job, with 334 test failures: https://jenkins.plone.org/job/plone-6.0-python-3.11/470/
I tried a few of the failing tests locally, but they passed.

This is after the merge of a PR from me:
https://github.com/plone/plone.app.linkintegrity/pull/87
The PR tests passed, so it could be coincidence.  Let's try.

PR #847, opened 2023-03-09
Checkout plone.app.linkintegrity again.

mauritsvanrees:
With the checkout, Jenkins was broken on all Plone 6.0 jobs, with 334 failures. Reverting that in https://github.com/plone/buildout.coredev/pull/846 helped: Jenkins is green again. So if I check it out again, the PR jobs should fail.  But they were green on the initial PR in linkintegrity. So maybe there is a subtle difference in how the core jobs and the PR jobs are run, like using multiple threads.

### DIFF
--- a/checkouts.cfg
+++ b/checkouts.cfg
@@ -31,8 +31,7 @@ auto-checkout =
     plone.app.portlets
     plone.base
     plone.app.layout
-# This may have broken Jenkins, but not sure yet.
-#    plone.app.linkintegrity
+    plone.app.linkintegrity
     Products.CMFPlacefulWorkflow
     plone.subrequest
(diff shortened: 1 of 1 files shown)


mauritsvanrees:
I started a single [PR job on 3.11](https://jenkins.plone.org/job/pull-request-6.0-3.11/299/).

mauritsvanrees:
So the PR job passed.
Now I wonder if the core jobs fail again when I merge this.  Let's see.
Anyone should feel free to revert immediately if you see that this again results in hundreds of test failures.

mauritsvanrees:
The core jobs fail massively again.  I am reverting.

PR #848, opened 2023-03-13
Revert plone.z3cform checkout

gforcada:
Seems that this breaks jenkins...


gforcada:

PR #849, opened 2023-03-15
Update various 6.0 versions, 2023-03-15

mauritsvanrees:
This includes several major releases of Zope packages: they drop support for older Python versions that we do not care about in Plone 6.

I have a small doubt about updating `importlib-metadata` from 5.20 to 6.0.0.  See its [history](https://importlib-metadata.readthedocs.io/en/latest/history.html#v6-0-0):

> [#419](https://github.com/python/importlib_metadata/issues/419): Declared Distribution as an abstract class, enforcing definition of abstract methods in instantiated subclasses. It’s no longer possible to instantiate a Distribution or any subclasses unless they define the abstract methods.
> Please comment in the issue if this change breaks any projects. This change will likely be rolled back if it causes significant disruption.

Only a few reactions were added, so should be okay.


mauritsvanrees:

jensens (approved):

PR #850, opened 2023-03-21
Removes pinning of collective.js.jqueryui in Plone 5.2

wesleybl:


mauritsvanrees:
We have kept this version pin until Plone 5.2.10.  I see no reason to drop it now, this late in the game.  I would rather just update the version to 2.2.0.

wesleybl:
Ok. But not pinn 2.2.0, because of: https://github.com/collective/collective.js.jqueryui/issues/53

PR #851, opened 2023-03-23
Fix link to GitHub Team, Developers

stevepiercy:
People are still wandering into v5 docs, and this change will need a rebuild. @polyester

Refs: https://github.com/plone/documentation/pull/1485

### DIFF
--- a/docs/agreement.rst
+++ b/docs/agreement.rst
@@ -60,7 +60,7 @@ Dealing with pull requests on GitHub
 
 Before we can merge a pull request, we have to check that the author has signed the contributor's agreement.
 
-If they're listed in https://github.com/plone?tab=members, the author has signed so we can go ahead and merge.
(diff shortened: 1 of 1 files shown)


davisagli (commented):
Now it's necessary to check both https://github.com/orgs/plone/teams/contributors/members and https://github.com/orgs/plone/teams/developers/members, right?

stevepiercy:
@davisagli yup, updated with current policy. One more review, if you please?

davisagli (approved):
It makes me wonder if we should add everyone from the Developers team to the Contributors team as well, so there's only one place to check. Thoughts, @fredvd?

Anyway, I think this is accurate for describing the status quo.

stevepiercy:
Sorry, one more review, please. I forgot this is in reStructuredText and I cleaned up some wording to complete this section.

davisagli (approved):

PR #852, opened 2023-03-28
Use latest pip 23, setuptools 67, wheel 0.40

mauritsvanrees:


mauritsvanrees:

mauritsvanrees:
Note that this *does* pollute the instance log with more deprecation warnings, like this:

```
DeprecationWarning: pkg_resources is deprecated as an API

DeprecationWarning: Deprecated call to `pkg_resources.declare_namespace('plone')`.
Implementing implicit namespace packages (as specified in PEP 420) is preferred
  to `pkg_resources.declare_namespace`.
```

See https://setuptools.pypa.io/en/latest/references/keywords.html#keyword-namespace-packages and https://github.com/zopefoundation/meta/issues/194

gforcada (approved):

PR #854, opened 2023-04-21
cleanup: remove Products.validation

gforcada:
It is no longer used by Plone core.

See https://github.com/plone/Products.validation/issues/9

I thought of leaving the constrain, but as anyway there are no new releases, the latest one is still safe to use 🍀


mauritsvanrees:
Since `collective.easyform` still uses this, I would say: leave it where it is in `versions.cfg` for the 6.0 series.

In 6.1 we could either move it to `versions-ecosystem.cfg`, or drop it altogether, especially if we move the relevant parts to `collective.easyform`.

jensens:
Easier to start over with a fresh one than rebasing here, see #933

PR #861, opened 2023-05-16
Add sanity check for bootstrap-pyenv.sh

pgrunewald:
I unfortunately called this script without having the pyenv script present and ended up pip installing everything on the default installation.

Adding here a sanity check to prevent this in the future.

### DIFF
--- a/bootstrap-pyenv.sh
+++ b/bootstrap-pyenv.sh
@@ -3,6 +3,11 @@
 PY="3.11.2"
 VENVBASE="plone-coredev-6.1"
 
+if ! which pyenv &> /dev/null; then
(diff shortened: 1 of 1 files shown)


pgrunewald:
I withdraw my PR, since it is still broken - at least for me. The commands do not work, but create various virtualenvs. It should be deleted, since the pyenv script is outdated anyway (displaying a warning, whenever used).

bootstrap.sh works as expected, though.

davisagli:
@pgrunewald Thanks for trying to improve the situation.

I think we should remove the bootstrap-pyenv.sh file, and add a few lines to the readme to say how to get started, rather than just linking to the docs which unfortunately still need a significant round of revisions.

PR #866, opened 2023-06-16
Take over coredev 6.0 changes in 6.1.

mauritsvanrees:
Remaining differences: plone.staticresources, plonetheme.barceloneta, mockup versions/sources. And let's use latest Products.CMFCore 3.1 here.
And bootstrap.sh: use Python 3.11 by default.


mauritsvanrees:

PR #868, opened 2023-06-29
Add playwright based robotframework tests

gforcada:
__Big disclaimer:__ all credit goes to @datakurre , I'm merely creating the PRs and doing the integration of the code 😅 

Part of https://github.com/plone/Products.CMFPlone/issues/3813

PR #869, opened 2023-06-29
Add version pins for robotframework-browser

gforcada:
With these version pins we can test the new changes in plone.app.robotframework to add support for playwright based tests (that uses robotframework-browser as a glue between playwright and robotframework).

That's part of #868 but it's the minimal changes needed so we can get https://github.com/plone/plone.app.robotframework/pull/149 to run.


gforcada:

gforcada:
@mauritsvanrees I'm merging it, so we can unblock https://github.com/plone/plone.app.robotframework/pull/149 in jenkins, it will fail, but not at running buildout 😅

mauritsvanrees (commented):
A few extra pins are fine.

PR #873, opened 2023-07-11
6.0: Raise plone.app.widgets version to 5.0.0 with deprecation warnings.

thet:
plone.app.widgets 5.0.0 deprecated itself, introduces deprecation warnings and has all code removed, which was moved to plone.app.z3cform in 4.1.0.

We are not sure if this should go into Plone 6.0 or should only be in 6.1+.
But plone.app.z3cform 4.1+ is in Plone since 6.0.4, which effectively does not use plone.app.widgets anymore.

So, should this PR be included in Plone 6.0?

/cc @mauritsvanrees @gforcada @jensens @petschki


petschki:
There's https://github.com/plone/Products.CMFPlone/pull/3686 which will remove p.a.widget dependency of CMFPlone. We will discuss possible dependency issues with @jensens this week here at arTSPrINT Vienna and as soon as that is merged there is no p.a.widgets dendency in core anymore. The 5.0 version is ment for backwards compatibility for addons only which should IMHO be shipped in Plone 6.1 but removed in Plone 7 (as the p.a.widgets deprecation warning is already saying)

gforcada (commented):
LGTM, but I will leave the decision to @mauritsvanrees although with the plans you made it seems to be 100% safe for this to get merged 😃 

Great job! 👍🏾

mauritsvanrees (approved):
Slowly getting back to community work after a vacation.

Seems fine.

PR #874, opened 2023-07-11
6.1: Raise plone.app.widgets version to 5.0.0 with deprecation warnings.

thet:
Same as https://github.com/plone/buildout.coredev/pull/873 but for Plone 6.1. Should be safe to be merged.

/cc @mauritsvanrees @gforcada @jensens @petschki


gforcada:
Great! I'm a bit confused though:

If the changes on 5.0.0 are deprecating the package, but still providing deprecation warnings (thus, you can still import `plone.app.widgets.foo` and it will be working) then a major version would not be needed, right?

Then when the deprecations are actually removed, then a major release would be needed.

Still, as the package is deprecated, and thus no new release is to be expected (?) a final breaking release might be a good idea 👍🏾 

Sorry for the rant 😆 😅

gforcada (approved):

thet:
@gforcada yeah, sorry. The release was made a bit quick, just after having a short discussion with @petschki . We realized that a minor release would have been enough.

PR #883, opened 2023-07-21
Raise z3c.form version in constraints.txt.

thet:


petschki:
I think the `constraints.txt` is generated when @mauritsvanrees makes a release ... btw: there's a preliminary https://dist.plone.org/release/6.1-dev/constraints.txt which already has z3c.form 5.1 ... this is used in the p.a.z3cform cleanup PR

thet:
Yes, true... I've seen that right after this careless merge :| At least it doesn't hurt.

mauritsvanrees:
Indeed, the `constraints.txt` in coredev and on dist.plone.org only get updated when I do something.  There are plans to change this, see https://github.com/plone/plone.releaser/issues/53 plus contained links.

But: no problem if you change this in coredev.

PR #890, opened 2023-10-04
Add support for Python 3.12

pbauer:


gforcada:
@pbauer do we want to add 3.12 support to Plone 6.0 first? 🤔 the jobs I created on Jenkins are for Plone 6.1, but adding them for Plone 6.0 should not be a problem 👍🏾

gforcada:
@pbauer `robotframework-browser` has already a 3.12 compatible release 🎉 I updated this PR to be on top of `6.0` branch, so we get `plone.api` and `plone.restapi` fixed `master` branches. This should make this PR pass 🍀

gforcada:
@mauritsvanrees the GHA `Create only instance and test with Buildout` fails due to `robotframework-browser` having a dependency on `click` but `click` version pin is only on `versions-extra.cfg`.

mauritsvanrees:
> @mauritsvanrees the GHA `Create only instance and test with Buildout` fails due to `robotframework-browser` having a dependency on `click` but `click` version pin is only on `versions-extra.cfg`.

This is exactly the kind of problem that this test is supposed to catch: `click` (and its dependencies) are now pulled in by Plone (or tests), so this version should be in `versions.cfg`.
I will move the pin.  Trying it out locally first.  Takes a while, as lots of packages need to be downloaded and/or compiled on 3.12 for the first time.

In 6.1 we could choose to use just one versions file.  But for 6.0 we should stick to what we have.

mauritsvanrees:
That worked.  Let's try Jenkins for good measure.

PR #897, opened 2023-10-31
Checkout Zope and use its master versions. [6.0]

mauritsvanrees:
We may need this for a dos_protection fix.


mauritsvanrees:

mauritsvanrees:
I have marked this as Draft PR, as we should not merge this just now.  But it is good to know that the tests pass.  If there is a new Zope release, we can switch to it without trouble.

PR #899, opened 2023-11-10
Add unofficial newer-versions.cfg to 5.2.

mauritsvanrees:
This file contains unofficial newer version pins for a few packages. These are known to work on Plone 5.2, but there are reasons to not include them officially.  The changes in there may be slightly more adventurous than is deemed wise at this late stage in the Plone 5.2 release cycle: currently (November 2023) Plone 5.2 is even out of maintenance support. Some do not work on Python 2, and we do not want to have two different version pins for core packages, to avoid confusion.

### DIFF
--- a/newer-versions.cfg
+++ b/newer-versions.cfg
@@ -1,0 +1,49 @@
+# This file contains unofficial newer version pins for a few packages.
+# These are known to work on Plone 5.2, but there are reasons to not include
+# them officially.  The changes in there may be slightly more "adventurous"
+# than is deemed wise at this late stage in the Plone 5.2 release cycle:
+# currently (November 2023) Plone 5.2 is even out of maintenance support.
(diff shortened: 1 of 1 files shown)


mauritsvanrees:
We don't need to run Jenkins: this file will not be used there.

ale-rt (approved):
LGTM :)
For Quaive we also have `plone.api = 2.0.5`

mamico (approved):

jensens (approved):

gforcada (approved):

mauritsvanrees:
> LGTM :) For Quaive we also have `plone.api = 2.0.5`

I would be hesitant to include that version here, as no tests are run for Plone 5.2.  I see Buildout files for 5.2 on latest master, but they seem unused.

Thanks for the reviews all, I will merge.

PR #902, opened 2023-12-07
New versions for 6.0, 2023-12-07

mauritsvanrees:
Let me draw your attention to one change: I have updated ZEO to 6.0.0.  This adds support for Python 3.12.  I have a small doubt about introducing the [other changes](https://pypi.org/project/ZEO/6.0.0/) in a Plone bugfix release. Especially "Switch to using async/await directly instead of @coroutine/yield" sounds like it could potentially have side effects that only become clear when it gets used.  So let's give this a chance on Python 3.12.  We can see later if we dare to introduce this for the other Python versions as well.

### DIFF
--- a/versions-extra.cfg
+++ b/versions-extra.cfg
@@ -3,8 +3,8 @@
 # that pull in ever more dependencies.
 # Note: version pins in this file can be removed at any time.
 [versions]
-argcomplete = 3.1.4
-argh = 0.29.4
+argcomplete = 3.1.6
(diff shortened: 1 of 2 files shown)


mauritsvanrees:

gforcada:
Sounds sensible, to only allow ZEO 6.0.0 on Python 3.12 👍🏾

gforcada (approved):

PR #903, opened 2023-12-09
Create mxdev sources and checkouts

thet:
Add scripts to create mxdev compatible ini files from sources.cfg and checkouts.cfg.

This feature is useful with the mxdev `include` feature, available since mxdev 3.1.0 and done at the Kitchen Sprint 2023 in Völs. See: https://github.com/mxstack/mxdev/pull/37

TODO
- [x] integrate it in tox release workflow.
- [ ] document somewhere
- [x] test a with mxdev

### DIFF
--- a/release/create-mxcheckouts.py
+++ b/release/create-mxcheckouts.py
@@ -1,0 +1,41 @@
+# -*- coding: utf-8 -*-
+"""Parse checkouts from a buildout configuration and write to a mxdev
+compatible ini file.
+
+This script is called by 'tox -c release/tox.ini -e create-mxcheckouts.py'
+
+Manual usage: bin/python create-mxcheckouts.py
+"""
+
+# From: create-constraints.py
+
+import os
+import sys
+
+from zc.buildout import buildout
+
+if not len(sys.argv) == 3:
+    print("ERROR. Usage: create-mxcheckouts.py buildout.cfg checkouts.ini")
+    sys.exit(1)
+config_file = sys.argv[1]
+config_file = os.path.realpath(config_file)
+checkouts_file = sys.argv[2]
+checkouts_file = os.path.realpath(os.path.join(os.getcwd(), checkouts_file))
+config = buildout.Buildout(config_file, [])
+
+checkouts = config.get("buildout", {}).get("auto-checkout", "").split("\n")
+
+with open(checkouts_file, "w") as cfile:
+    cfile.write(f"# File created by {os.path.relpath(__file__)}\n")
+    cfile.write(f"# Parsed from {os.path.relpath(config_file)}\n")
+
+    # Set default to not checkout any sources.
+    cfile.write("\n[settings]\n")
+    cfile.write("default-use = false\n")
+
+    cfile.write("\n")
+    for name in sorted(checkouts, key=lambda x: x.lower()):
+        cfile.write(f"\n[{name}]\n")
+        cfile.write("use = true\n")
+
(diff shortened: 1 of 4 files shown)


mauritsvanrees (commented):
Interesting, thanks.

Have you seen the work I did on `plone.releaser` the past months?  For example in https://github.com/plone/plone.releaser/pull/62 I added a `bin/manage versions2constraints` command and recently used this on the coredev 6.1 branch.
There is code to read on the one hand a `checkouts.cfg` and `sources.cfg` and on the other hand a `mxdev` ini file.  Just the translation between the two needs to be hooked up.
And I obviously did not yet use the new `include` functionality.

My preferred solution would be:

* Use the code that already is in `plone.releaser`, as this has tests.
* Improve it so it can use the new `include` functionality from `mxdev`.
* Add a `bin/manage buildout2mxdev` (or `buildout2pip`) command that translates all Buildout versions, checkouts and sources files to pip/mxdev files.

We can consider using your code for now, as it is available and seems to be working.  Later we could switch to using `plone.releaser`.

Well, the tox command currently fails for me:

```
$ tox -c release/tox.ini -e mxdev
mxdev: commands[0] /Users/maurits/community/plone-coredev/6.1/release> python create-mxsources.py /Users/maurits/community/plone-coredev/6.1/release/../buildout.cfg /Users/maurits/community/plone-coredev/6.1/release/../sources.ini
  File "/Users/maurits/community/plone-coredev/6.1/release/create-mxsources.py", line 15, in <module>
    from zc.buildout import buildout
ModuleNotFoundError: No module named 'zc'
```

mauritsvanrees:
BTW, do you and @jensens indeed prefer to split the sources and checkouts in two files?  I would be fine with either one or two files, but two files make it a bit easier to translate between our current buildout config and mxdev, so I would be happy with that approach.  A short `checkouts.ini` makes it easier to see which packages are checked out.

BTW, several commands already work on both buildout and mxdev files.  If you temporarily rename or copy `checkouts.ini` to `mxdev.ini`, you get this:

```
$ bin/manage check-checkout Products.CMFPlone
YES, your package Products.CMFPlone is on auto checkout in mxdev.ini.
YES, your package Products.CMFPlone is on auto checkout in checkouts.cfg.
```

jensens:
I think having 3 files has it advantages.
- `mx.ini` over all setup
- `mxsources.ini` (yes prefix, so it stands together) 
- `mxcheckout.ini` with only the checkouts.

This way one may use only sources with own checkouts/branches(think PLIPS) overrides.

release/create-mxcheckouts.py
```diff
@@ -0,0 +1,41 @@
...
+import os
+import sys
+
+from zc.buildout import buildout
+
+if not len(sys.argv) == 3:
+    print("ERROR. Usage: create-mxcheckouts.py buildout.cfg checkouts.ini")
+    sys.exit(1)
+config_file = sys.argv[1]
+config_file = os.path.realpath(config_file)
```
gforcada, line 21: Using pathlib.Path makes it all much nicer to read 😃

Sorry on the phone ,I can provide suggestions later 😃

gforcada (commented):
Looks good, with @mauritsvanrees plone.releaser changes we are getting closer and closer to replace zc.buildout finally 😃

mauritsvanrees:
Closing this in favour of https://github.com/plone/buildout.coredev/pull/945.

PR #906, opened 2024-01-29
Upgrade Pillow to 10.2.0.

mauritsvanrees:
This should work, now that https://github.com/plone/plone.scale/pull/90 is merged.

I am inclined to say we only do this for Plone 6.1, not 6.0.  The release has backwards incompatibilities.  Should not be noticeable under normal use, but you never know what projects have as extra code. I am not aware of anything new in Pillow 10 that we would really want to actively use in Plone.  So 9.5 should still be fine for Plone 6.0.


mauritsvanrees:

davisagli (approved):

wesleybl:
@mauritsvanrees would it be possible to use `Pillow` version 10 in Plone 6.0? In theory, version 9.5.0 does not support Python 3.12:

https://github.com/python-pillow/Pillow/blob/9.5.0/setup.cfg#L19

And we don't have wheels for Python 3.12 for version 9.5.0:

https://pypi.org/project/pillow/9.5.0/#files

PR #908, opened 2024-02-10
Update versions (6.0).

mauritsvanrees:
Includes latest pip and setuptools.
Also, we were using ZEO 6.0.0 only on Python 3.12 until now (Plone 6.0.9), but now we use it for all Python versions.


mauritsvanrees:

davisagli (approved):

PR #910, opened 2024-02-23
Uses TZ = UTC in buildout - 6.0

wesleybl:
This ensures that tests using datetime will work locally on machines using TZ = GMT

It's the same thing that is in Plone 5.2 but was lost in Plone 6:

https://github.com/plone/buildout.coredev/blob/5.2/tests.cfg#L150-L158


wesleybl:

PR #920, opened 2024-04-23
Checkout Zope and use its versions.cfg for testing.

mauritsvanrees:
Some recent Zope changes may require changes in Plone.  Let's see what breaks.

Locally I already saw test failures in `plone.dexterity`, which should be easy to fix either in the tests or in Zope.

### DIFF
--- a/checkouts.cfg
+++ b/checkouts.cfg
@@ -10,6 +10,7 @@ auto-checkout =
     plone.app.upgrade
     Products.CMFPlone
 # These packages are manually added, or automatically added by mr.roboto:
+    Zope
     plone.restapi
     plone.volto
     plone.app.contenttypes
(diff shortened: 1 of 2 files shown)


mauritsvanrees:
Okay, GitHub Actions has problems with MacOS, which is the case on the main 6.0 and 6.1 branches as well, so ignore that.

mauritsvanrees:
One unrelated failing robot test on one of the two Jenkins jobs.  Seems good there.

I rebase on 6.0 to get my latest gh-actions fix in.

mauritsvanrees:
Everything works.  So current Zope master and its versions should be safe to use.
But let's not do that in the Plone releases of this week, it can use more testing.

PR #921, opened 2024-04-23
Update to lxml 5.2.1.

mauritsvanrees:
* We are using version 4 until now.
* Version 5 gives a test error in Diazo.  That is fixed in https://github.com/plone/diazo/pull/88.
* In lxml 5.2.0 and higher, `lxml.html.clean` was factored out into an extra package, and not pulled in by default.  Fixed by https://github.com/plone/Products.PortalTransforms/pull/62.
* This coredev PR for 6.0 checks out branches for those two packages, it updates lxml to latest, and it adds a pin for the new `lxml_html_cleaner` package.

Once green, this PR should not be merged, because we do not want to use those PR branches. The right order would be something like this:

* Merge the Diazo PR, it is fine for lxml 4 and 5.
* Add `lxml-html-clean = 0.1.1` to the versions in 6.0 and 6.1.
* Merge the PortalTransforms PR, which then pulls in the new package without giving an error for a missing version pin.
* Update the lxml version to latest 5.2.1.


mauritsvanrees:

jensens:
A long awaited update of lxml, nice! Thanks @mauritsvanrees

mauritsvanrees:
Trying again after a fix in PortalTransforms:

mauritsvanrees:
One unrelated test failure in `plone.app.caching.tests.test_operation_utils.ResponseModificationHelpersTest.test_cacheInBrowserAndProxy_minimal` in one of the Jenkins jobs.  Looks like we are good to go.

I will do the changes in the order that I outlined in the original comment.

mauritsvanrees:
All done, on 6.0 and 6.1.

PR #931, opened 2024-05-15
POC: use mxdev

mauritsvanrees:
This uses the `bin/manage buildout2pip` command from https://github.com/plone/plone.releaser/pull/72 to create `mxsources.txt` and `mxcheckouts.txt` from the Buildout sources and checkouts.  Then we add some more files to use this, and update the Makefile:

```
$ make help
clean                          Remove old virtualenv and creates a new one
help                           This help message
mxdev-generate                 mxdev: generate requirements and constraints files without pulling sources
mxdev-update                   mxdev: update requirements and constraints files and pull the sources
pip-bootstrap                  Pip: Bootstrap a venv for tests (future: several venvs with less installed)
pip-test                       Pip: Run only a few unit tests, as proof of concept.
pip-update                     Pip: Update a venv for tests (future: several venvs with less installed)
test-acceptance                Run acceptance tests
test                           Run tests
```

The Makefile targets starting with `pip` or `mxdev` are new.

Usage would be:

```
# Create/refresh the mxdev sources and checkouts (you need to use the `plone.releaser` branch from the linked PR (I committed the result):
bin/manage buildout2pip
# Create venvs/test/ installing `uv` and `mxdev`:
make pip-bootstrap
# Use mxdev to (re)generate requirements and constraints files:
make mxdev-generate
# Instead, you can call `make mxdev-update` to do the same plus pull the packages that should be checked out.
# Install all packages, which includes test packages: 
make pip-update
# Run the real unit tests (without any layers) of CMFPlone:
make pip-test
```

Some points of interest:

* `make pip-test` will need lots of changes before it can be a replacement of `bin/test`.
* Bug in `mxdev`: it does not comment out `Plone` and `Products.CMFPlone` from the constraints, probably because they have capital letters.  I made a manual change in `constraints-mxdiv.txt` to work around this.  I created an [issue in `mxdev`](https://github.com/mxstack/mxdev/issues/45).
* At some point I would want multiple venvs without the test dependencies: one with only `pip install Products.CMFPlone`, with with `Plone`, one with `plone.volto`, and one with `plone.classicui`.  Tricky is that the `Plone` package is always in the checkouts, so we would always have `-e src/Plone`, making it hard to have a venv with only `Products.CMFPlone`.  But the single venv with everything including test dependencies seems fine as a start.
* We do not generate any wsgi and `zope.conf` files, like cookieplone or other generators would do.  That is for later.  But maybe the ones generated by Buildout are fine for now.


mauritsvanrees:
Closing this in favour of https://github.com/plone/buildout.coredev/pull/945.

PR #936, opened 2024-06-19
Distributions PLIP config: add zeo with 5 clients.

mauritsvanrees:
One zeoclient for each of: Plone, Products.CMFPlone, plone.distribution, plone.volto, plone.classicui. So only one of those packages in the eggs.
Then you can run 'bin/zeoserver start' and then start however much of those zeoclients you want. Port numbers between 8080 and 8084.

### DIFF
--- a/plips/plip-distributions.cfg
+++ b/plips/plip-distributions.cfg
@@ -18,6 +18,7 @@ auto-checkout +=
     plone.exportimport
     plone.classicui
     plone.volto
+    plone.app.testing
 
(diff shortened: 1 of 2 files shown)


mauritsvanrees:
I have added `plone.app.testing` to the checkouts, and use branch `distributions` for that package plus `Products.CMFPlone`.  I will add some info in https://github.com/plone/Products.CMFPlone/issues/3961

I will just merge this PR, it only affects the PLIP config.

PR #937, opened 2024-06-19
Fix typo in a comment in `checkouts.cfg`

stevepiercy:
For something this trivial, fixing a typo in a comment, does it need a formal change log entry?


mauritsvanrees (approved):
We do no changelog entries in the `buildout.coredev` repo.

Thanks for the fix.  I must have overlooked that thousands of times. :-)

PR #938, opened 2024-07-08
Set https://dist.plone.org/release/6.1-dev/ in find-links.

mauritsvanrees:
Otherwise we may not find the Products.CMFPlone 6.1.0a4.dev0 that I put there.

Set Products.CMFPlone 6.1.0a4.dev0 in versions, though we don't really use it: we always check it out.


mauritsvanrees:
I am mostly away from keyboard, so please someone merge when green.

gforcada (approved):

PR #942, opened 2024-08-05
pin docutils 0.21.2

stevepiercy:
See https://github.com/plone/plone.restapi/pull/1798#issuecomment-2270076053

### DIFF
--- a/versions.cfg
+++ b/versions.cfg
@@ -22,6 +22,7 @@ zc.buildout = 3.0.1
 nt-svcutils = 2.13.0
 
 # OVERRIDES
+docutils==0.21.2
 
 # CORE PLONE.
 # These packages are what you get when installing Plone plus test dependencies,
--- a/versions.cfg
+++ b/versions.cfg
@@ -22,7 +22,7 @@ zc.buildout = 3.0.1
 nt-svcutils = 2.13.0
 
 # OVERRIDES
-docutils==0.21.2
+docutils = 0.21.2
 
 # CORE PLONE.
 # These packages are what you get when installing Plone plus test dependencies,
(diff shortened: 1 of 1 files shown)


versions.cfg
```diff
@@ -22,6 +22,7 @@ zc.buildout = 3.0.1
 nt-svcutils = 2.13.0
 
 # OVERRIDES
+docutils = 0.21.2
```
davisagli, line 26:
```suggestion
docutils = 0.21.2  # Zope uses an older version for its Sphinx theme, which Plone does not use
```

versions.cfg
```diff
@@ -22,6 +22,7 @@ zc.buildout = 3.0.1
 nt-svcutils = 2.13.0
 
 # OVERRIDES
+docutils = 0.21.2  # Zope uses an older version for its Sphinx theme, which Plone does not use
```
davisagli: Sorry, this comment syntax doesn't seem to work on the same line.

```suggestion
# Zope uses an older version for its Sphinx theme, which Plone does not use
docutils = 0.21.2
```

mauritsvanrees:
In coredev 6.1 we try to keep `versions.cfg` and `constraints.txt` in sync all the time, which mostly works automatically, except when you update a version by hand.
I have fixed it in commit 1e30099a1d0a428d9114c57d59fafbe34f57fba1

(In 6.0 we do not do this, there I sync the versions to the constraints when I make a Plone release.)

davisagli:
@mauritsvanrees Thanks, I didn't know they both need to be updated manually at the same time. It would be helpful to have a comment at the top to give that reminder.

PR #932, opened 2024-05-17
Enhance with proper Makefile

jensens:
- goes into #931 
- use mxmake based Makefile
- all in one step

 goal: `make zope-start` does all in one: 
- venv, 
- install regenerating mx[sources|checkouts].txt (ini), 
- run mxdev 
- run install
- create instance files
- run zope

PR #945, opened 2024-08-29
Add mxmake and mxdev for creating an instance and test using pip.

mauritsvanrees:
This is part of https://github.com/plone/Products.CMFPlone/issues/3670

This combines three PRs:

* https://github.com/plone/buildout.coredev/pull/903
* https://github.com/plone/buildout.coredev/pull/931
* https://github.com/plone/buildout.coredev/pull/932

See `README-make.md` for instructions.

This actually needs a change in `Products.CMFPlone`: it needs a `five:registerPackage`. Otherwise the Zope instance only works with a release of Products.CMFPlone, and not with an editable install (or src checkout). I will make a [PR](https://github.com/plone/Products.CMFPlone/pull/4003) for that.

There is stuff to improve, especially running the tests, but I would be in favour of merging this soonish so people can try it out.

If you don't use the Makefile and the other new files, then nothing changes: you can keep using Buildout as you were used to.

@gforcada With this in place, it would be good if `mr.roboto` calls `bin/manage buildout2pip` when it changes anything, especially when it adds a checkout.  Or it could call `bin/manage add-checkout <package name>` and let `plone.releaser` handle it.  But I see I would need to fix that command first.  PR: https://github.com/plone/plone.releaser/pull/74


mauritsvanrees:
The changes should have no influence on Jenkins, but let's run the tests to be sure.

mxtests.ini
```diff
@@ -0,0 +1,125 @@
...
+[plone.app.layout]
+extras = test
+
+[plone.app.linkintegrity]
+extras = test
+
+[plone.app.portlets]
+extras = test
+
+[#    plone.app.robotframework]
```
gforcada: typo? 🤔

gforcada (commented):
All looks good, I can basically not approve it as I'm still newbie on the `mxmake` stack 👶🏾 

But all in all, specially the README is really helpful to navigate this new tooling 👍🏾 ✨

mxtests.ini
mauritsvanrees: Ah, well spotted.  I did this with a script or some bash line (it was a while ago) with `tests.cfg` as input.  This line in `tests.cfg` thwarted that script:

```
#    plone.app.robotframework [test]
```

Fixed by properly commenting it out in `mxtests.ini` as well.

mauritsvanrees:
Rebased and updated on current 6.1 branch.

mauritsvanrees:
I have merged it.  Let's start using it, and improve where needed.

PR #948, opened 2024-09-16
Update versions for `robotframework-browser = 18.8.1`

petschki:


mauritsvanrees (approved):
Thanks.  I have added two small commits, removing the `grpcio-tools` pin and two version annotations.  See commit messages.

petschki:
`rfbrowser init` on Jenkins warns that node version is outdated and playwright installation has missing dependencies. See https://jenkins.plone.org/job/plone-6.1-python-3.12-robot-chrome/552/console ... problem?

gforcada:
I updated the node version to `v20.5.1` is that enough? 🤔

davisagli:
@petschki @mauritsvanrees It seems like the robot tests have been much more flaky on Jenkins since this was merged.

PR #950, opened 2024-09-20
Update to zc.buildout 3.1.1 and setuptools 75.1.0.

mauritsvanrees:
Note that setuptools 75 removes the deprecated 'upload' and 'register' commands from 'setup.py', but we should not need that. At least zest.releaser calls 'twine' to upload or register to PyPI.


mauritsvanrees:

PR #955, opened 2024-10-08
Move some pins from versions-extra.cfg to versions.cfg.

mauritsvanrees:
Otherwise running `bin/buildout -c bare.cfg` fails after the recent change, because some packages pull in `zest.releaser` in their `test` extra.
This also updates a few of those pins.

We may want to rethink whether we still want this distinction in three versions files, or if we should make a different distinction.


mauritsvanrees:

davisagli:
> some packages pull in zest.releaser in their test extra

That sounds wrong, if it isn't actually needed to run the tests.

On the other hand, I don't see much reason to keep 2 separate lists of versions. They are constraints to be used _if_ the packages are installed. We should have one set of supported versions, not two.

PR #956, opened 2024-10-11
Checkout Zope and try Python 3.13. [Don't merge]

mauritsvanrees:
Zope has [declared Python 3.13 compatibility](https://github.com/zopefoundation/Zope/pull/1230) so let's see how Plone fares.


mauritsvanrees:
I already see [one error](https://github.com/plone/buildout.coredev/actions/runs/11288553704/job/31396554755?pr=956) due to some new version pin:

```
File "...plone.base-2.0.1-py3.10.egg/plone/base/tests/messages.rst", line 39, in messages.rst
Failed example:
    msg.mapping
Expected:
    {'name': 'Plone'}
Got:
    mappingproxy({'name': 'Plone'})
```

mauritsvanrees:

mauritsvanrees:
Some newer pins are needed for Python 3.13, mostly so that binary wheels are available instead of needing to compile source distributions for packages with C/Rust code.  But in general I want to upgrade versions again, though the robotframework ones may be tricky again.

First:

* That test failure on Python 3.10 in `plone.base` should be fixed. Seems just an easy fix in the test itself at first glance.
* And Zope 5.11 is actually out now, so we should use that.

Those two can be done first.  Python 3.13 can happen later.  It was just a quick try.
Moving on to other work for the moment.

mauritsvanrees:
We are using the released Zope 5.11 now.

PR #957, opened 2024-10-18
Update robotframework-browser versions [6.0]

petschki:
Since node on jenkins is now correctly configured we can update the rfbrowser versions again.


petschki:

PR #965, opened 2024-10-28
Create a `ci.cfg` for minimal CI config

petschki:
Note: I've temporary tricked jenkins PR-6.1 jobs config in order to test this:

```bash
BUILDOUT_CFG="$([[ -s 'ci.cfg' ]] && echo 'ci.cfg' || echo 'buildout.cfg')"
buildout buildout:git-clone-depth=1 -c $BUILDOUT_CFG
```

### DIFF
--- a/ci.cfg
+++ b/ci.cfg
@@ -1,0 +1,6 @@
+[buildout]
+extends =
+    bare.cfg
+auto-checkout -=
+    docs
+    mockup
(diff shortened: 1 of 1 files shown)


mauritsvanrees (approved):
So this avoids installing all kinds of parts that are not used on Jenkins.  And it avoids the docs and mockup checkout that are not used there.  That makes sense.

What do you think @gforcada.  Are we overlooking anything?

It probably makes sense to move the `robot` part into `bare.cfg`.

gforcada:
Looks good to me 👍🏾

PR #969, opened 2024-11-30
Dev containers support for VSCode

xiru:
Add initial support for dev containers on VS Code.

### DIFF
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -1,0 +1,18 @@
+// For format details, see https://aka.ms/devcontainer.json. For config options, see the
+// README at: https://github.com/devcontainers/templates/tree/main/src/python
+{
+	"name": "Plone 6.1",
+	"image": "mcr.microsoft.com/devcontainers/python:1-3.12-bullseye"
+
+	// Use 'portsAttributes' to set default properties for specific forwarded ports. 
+	// More info: https://containers.dev/implementors/json_reference/#port-attributes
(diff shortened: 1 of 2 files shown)


.devcontainer/devcontainer.json
```diff
@@ -0,0 +1,18 @@
...
+	// More info: https://containers.dev/implementors/json_reference/#port-attributes
+	"portsAttributes": {
+		"8080": {
+			"label": "Plone Backend",
+			"onAutoForward": "notify"
+		}
+	},
+
+	// Use 'postCreateCommand' to run commands after the container is created.
+	"postCreateCommand": "pip3 install --user -r requirements.txt"
```
davisagli: @xiru We should run `make install` here. This takes care of some other steps that build the requirements file to include some packages from github (for the packages that have unreleased changes).

We'll need a base image that includes tools like make, git, and ssh.

davisagli (commented):
@xiru Thanks for running with this while I was discussing docs with the GSOC students. Here are a couple things I think we'll need to update. Come grab me if you want to discuss.

xiru:
@davisagli added a script to run the buildout as part of the dev container creation. Now we should be OK. Tested in my laptop and it worked as expected, environment was created, vc code sends a notification about the service running, you click it and opens a browset in you local machine, connecting localhost:8080 and it actually connects the 8080 port  inside the container.

davisagli:
Closing in favor of #970, which uses `make install` and fixes the virtual hosting

Issue #967, opened 2024-11-25
GHA: Must switch to macos-13


mauritsvanrees:
An action run on coredev just [failed](https://github.com/plone/buildout.coredev/actions/runs/12013410842/job/33486868272) with this:

"The macOS-12 environment is deprecated, consider switching to macOS-13, macOS-14 (macos-latest) or macOS-15. For more details, see https://github.com/actions/runner-images/issues/10721"

gforcada:
Seems this was done some time ago ✔️

PR #974, opened 2024-12-06
6.0: drop Python 3.8, drop macos-12, update Zope/pip/st/buildout.

mauritsvanrees:
Python 3.8 is out of security support, so I will officially drop it from Plone 6.0 soon.  As we say in the [release schedule](https://plone.org/download/release-schedule): "Plone 6.0 works on Python 3.8, but this Python version reaches end of life in October 2024. At that point, Plone 6 will drop support for Python 3.8."  See also https://github.com/plone/jenkins.plone.org/pull/377

We also need to stop testing on macos-12 as it may fail soon.  See https://github.com/plone/buildout.coredev/issues/967 Updated to macos-13 for now.

Updated pip, setuptools, buildout.
For the moment I kept setuptools at the highest version that still supports Python 3.8, but I want to increase that.  It would break Jenkins currently without the other PR I mentioned.

Update Zope to latest 5.11.1.

I will make this a draft PR, as the `Products.validation` tests fail due to an updated `zope.i18nmessageid`.  I am preparing a workaround for that: https://github.com/plone/Products.validation/pull/14

And plone.base fails, which needs a back port:
https://github.com/plone/plone.base/pull/74

### DIFF
--- a/.github/workflows/bare.yml
+++ b/.github/workflows/bare.yml
@@ -8,7 +8,7 @@ jobs:
       matrix:
         python-version:
         # Use the oldest supported Python version, as that may pull in more versions.
-        - "3.8"
+        - "3.9"
     runs-on: ubuntu-latest
     steps:
     - name: locale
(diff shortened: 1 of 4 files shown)


mauritsvanrees:

PR #975, opened 2024-12-10
Update various versions for 6.0.

mauritsvanrees:
This takes over most of the suggestions from the [versions check artefact](https://github.com/plone/buildout.coredev/actions/runs/12264982756).

Note that this updates `Pillow` from 9.5.0 directly to latest 11.0.0.  I held version 10 back until now because of a few incompatibilities that may theoretically break for some people with custom code and special wishes.  But there are a few security fixes in the 10.x series, so we do want to keep this updated.  Might as well move on to 11 then.


mauritsvanrees:

gforcada (approved):

PR #979, opened 2025-01-22
Use Zope 5.12 [6.1]

mauritsvanrees:
If I see it correctly, the only differences are actually a new Zope and a new zope.interface.


mauritsvanrees:

PR #951, opened 2024-09-26
Use zc.buildout 3.2.0 to optionally extend local.cfg. [6.1]

mauritsvanrees:
I added this option in https://github.com/buildout/buildout/pull/665

For inspiration, here is part of my own `local.cfg`:

```
[buildout]
develop +=
    /Users/maurits/community/zest.releaser
auto-checkout +=
    plone.releaser
custom-eggs +=
    pdbpp

allow-picked-versions = true

[remotes]
plone = git@github.com:plone
collective = git@github.com:collective
zope = git@github.com:zopefoundation

[versions]
zest.releaser =
```

Originally the file started with these lines:

```
[buildout]
extends =
   buildout.cfg
```

and then I always ran `bin/buildout -c local.cfg`. When I then wanted to use `plips/plip-distributions.cfg`, it got tricky because this was in a different directory, so I needed actual changes in git. Now it works in both cases thanks to the `optional-extends`.

PR #1003, opened 2025-03-10
Optionally extend local.cfg if it exists. [6.0]

mauritsvanrees:
We already have this in 6.1 since [October](https://github.com/plone/buildout.coredev/pull/951), so also in 6.2. I would like to have it the same in all Plone 6 versions.

cc @petschki because I think you are one of the people who may already be using a `local.cfg` in 6.0, so you would need to slightly change that file locally.

### DIFF
--- a/core.cfg
+++ b/core.cfg
@@ -4,6 +4,10 @@ extends =
     versions-extra.cfg
     bare.cfg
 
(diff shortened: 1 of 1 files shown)


petschki:
Sure 🙌🏼
thanks for the hint

PR #1007, opened 2025-04-01
Update to twine 6.1.0

davisagli:
Setuptools > 77 generates sdists with a PKG-INFO file using Metadata-Version: 2.4

Older versions of twine don't recognize this metadata version, which can lead to the following error while running twine check (the "missing fields" are actually there, but not found because of the unknown Metadata-Version):
```
ERROR    InvalidDistribution: Metadata is missing required fields: Name,        
         Version.                                                               
         Make sure the distribution includes the files where those fields are   
         specified, and is using a supported Metadata-Version: 1.0, 1.1, 1.2,   
         2.0, 2.1, 2.2, 2.3.                                                    
```

This updates to the latest version of twine which recognizes Metadata-Version 2.4.

I'll make PRs for the 6.0 and 6.1 branches of buildout.coredev as well. @mauritsvanrees How do we update the -dev constraints on dist.plone.org?


gforcada (approved):

mauritsvanrees (approved):
Thanks!

I will update dist.plone.org today or tomorrow.  Almost bed time here.

mauritsvanrees:
I have updated the versions and constraints of 6.0-dev, 6.1-dev, 6.2-dev on dist.plone.org.

PR #1012, opened 2025-04-08
Use latest zc.recipe.egg, zc.buildout, setuptools. [6.1]

mauritsvanrees:

PR #1013, opened 2025-04-08
Use latest zc.recipe.egg, zc.buildout, setuptools. [6.2]

mauritsvanrees:


mauritsvanrees:
Strange.  This fails with:

```
Error: Couldn't find a distribution for 'five.customerize==4.0'.
```

or for the ecosystem tests:

```
Error: Couldn't find a distribution for 'plone.app.mosaic==3.2.3'.
```

The same [PR for 6.1](https://github.com/plone/buildout.coredev/pull/1012) works fine.

The `five.customerize` error could be because this is a native namespace, which differs from Plone 6.1.  But for `plone.app.mosaic` the version is the same.

I want to see if Jenkins has the same problem:

mauritsvanrees:
PyPI seems to be having some problems.  Locally when I disable my eggs cache, I get the same error getting `plone.app.iterate`.  It goes wrong when getting the index page: 'HTTP Error 503: Service Unavailable'.  But Buildout does not report this.  This is actually in a part of setuptools that we override, and the original method (`PackageIndex.process_url`) does not report this either: only an authentication error gets reported.

I will see about yet another PR for Buildout.

mauritsvanrees:
PyPI is recovering, I will restart GitHub Actions.

PR #985, opened 2025-02-14
Run tests from makefile

rnixx:
This PR contains the stub work for running the test suite via Makefile.

I added a mxsources_static.ini which shows how to configure tests for package Products.CMFPlone. mxsources.ini is generated and all changes are overwritten.

Also qa.test and qa.coverage topics from mxmake have been included in the makefile.

PR #1020, opened 2025-05-28
Support running tests with mxmake/pip-installed packages

mauritsvanrees:
Just try it: `make test`.  For the moment as proof of concept this only runs the unit tests.  But locally it works for the normal tests as well.  I did not try the robot tests, they may need some more setup with `rfbrowser`.

Main changes:

* Enable the `qa.test` domain in the `Makefile`, and run `mxmake update` to update some relates files.
* Add a manually written `run-tests.sh` script in the repo root, instead of the automatically generated `.mxmake/files/run-tests.sh`.
* Upgrade to fresh `zope.testrunner` 7.4 so we can use `--auto-path`.  This is needed, because otherwise we would need to add a `mxmake-test-path` option in each section in `mxsources.ini`, like was done as example in PR #985.

This PR also includes a minor sync of constraints and checkouts.

I wanted to add a GitHub Action job to run `make test`.  But this could use a change in `mxdev` so we do not make a full clone of each checkout, but use a git depth of 1.  I have [opened an issue](https://github.com/mxstack/mxdev/issues/50).  [Update: that PR is merged and released, and we use it now.]


mauritsvanrees:
This is now also tested on GitHub Actions, by running `make test` on Python 3.10-3.13 on Ubuntu and Mac.  

Windows did not quite work, giving several [test failures](https://github.com/plone/buildout.coredev/actions/runs/15449045805/job/43486236797?pr=1020).  This would likely happen when using `bin/buildout` and `bin/test -u` as well, so this should not be a problem.

So I think this is ready for merge.

Let's do Jenkins for the minor version updates:

mauritsvanrees:

mauritsvanrees:
For me this is fine to merge.  Then core developers can start using this more seriously, and later we could work in the direction of letting Jenkins use this instead of running buildout.

Most of this would be fine for the 6.1 branch as well, really, we can see about updating that.

mauritsvanrees:
We need this, I merge before it gets out of sync with the 6.2 branch again.  There is certainly room for improvement, but now we can improve incrementally.

PR #1021, opened 2025-06-06
Do not clone the Plone documentation. [6.1]

mauritsvanrees:
See https://github.com/plone/documentation/pull/1936#issuecomment-2946455205 and the comments after that.

### DIFF
--- a/checkouts.cfg
+++ b/checkouts.cfg
@@ -3,7 +3,6 @@ always-checkout = force
 # always-checkout = false
 auto-checkout =
 # These packages are always checked out:
-    docs
     mockup
     Plone
     plone.app.locales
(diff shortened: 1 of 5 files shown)


mauritsvanrees:
For this one and https://github.com/plone/buildout.coredev/pull/1022 I am running one test on Jenkins, just to check that there is nothing on Jenkins that is using the `docs-directory` setting for anything.

jensens (approved):
I think it is a good idea to not have the documentation in here. It's for the most part confusing.

davisagli (approved):
I think it makes sense to maintain tooling for the documentation in one place (https://github.com/plone/documentation/) and not 2.

PR #1026, opened 2025-06-11
Use the 5.x series of plone.app.theming for Plone 6.0.

thet:
In plone.app.theming I'd like to depend on plone.base > 3.0.0 to use the `is_truthy` utility function while Plone 6.0 has a dependency on plone.base < 2. This would be a breaking change in plone.app.theming and I need to cut a branch 5.x use it in Plone 6.0.

Here it is.


mauritsvanrees (approved):
Thanks.

PR #1028, opened 2025-06-17
Use latest pytest versions. [6.1]

mauritsvanrees:
Not really used by the coredev buildout AFAIK.
But when I make some local changes in the test setup of `plone.distribution` and `plone.classicui` to use these versions, their tests pass.


mauritsvanrees:
Probably not needed to try Jenkins, but let's do it for good measure.  I don't want to introduce incompatibilities this close to a 6.1.2 release.

mauritsvanrees:
Seems fine, I merge.

PR #1030, opened 2025-07-18
bump pyrepl to fix pdbpp on mac

pbauer:


mauritsvanrees:
@pbauer Can you do this in `versions.cfg` as well please?  Usually I would edit that file, and then run `bin/manage versions2constraints` to update the constraints.  (Some time this year I will add a command for the other way around.)

pbauer:
done, sorry for missing that.

PR #1031, opened 2025-08-17
Add version pin for python-magic [6.1]

gforcada:
This python distribution is needed in order to allow `Products.MimetypesRegistry` to use it.

See https://github.com/plone/Products.MimetypesRegistry/pull/20

`buildout` in `buildout.coredev` does not allow unpinned versions, so we have to pin it before we can test the above MR in Jenkins.

Note that this should also be ported to `6.0` and `6.2` branches, but first let's agree if we want to ever get the above MR merged, and then we can cherry-pick these changes in `buildout.coredev` to the other branches.

### DIFF
--- a/versions-extra.cfg
+++ b/versions-extra.cfg
@@ -36,6 +36,7 @@ PyGithub = 2.6.1
 PyNaCl = 1.5.0
 pyparsing = 3.2.3
 pyrepl = 0.11.4
+python-magic = 0.4.27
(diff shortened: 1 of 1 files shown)


davisagli:
I think it makes sense to try this and then see if https://github.com/plone/Products.MimetypesRegistry/pull/20 passes.

davisagli (approved):

PR #1002, opened 2025-03-10
Update to latest robotframework version

petschki:
This fixes "_start_keyword" errors

PR #1041, opened 2025-10-02
latest robotframework versions [6.2]

petschki:
~~see https://github.com/plone/buildout.coredev/pull/1002~~ unfortunately after merging, there was an error with the jenkins robotframework plugin again when generating the reports. Do not merge until we know whats going on there or a plugin update is available.

### UPDATE

robotsuite XML merger is fixed in `3.0.1`


versions.cfg
```diff
@@ -265,21 +264,21 @@ requests-toolbelt = 1.0.0
...
-# but https://github.com/robocorp/robotframework-lsp/issues/947
+robotframework = 7.3.2
 robotframework-assertion-engine = 3.0.3
-robotframework-browser = 19.3.1
+robotframework-browser = 19.8.0
 robotframework-debuglibrary = 2.5.0
-robotframework-lsp = 1.10.1
+robotframework-lsp = 1.13.0
 robotframework-pythonlibcore = 4.4.1
+# core tests do not use selenium anymore -> remove?
```
mauritsvanrees, line 273: For me it is fine to remove the selenium pins.

petschki: There are some addons (especially easyform) which rely on selenium and buildout for testsetup ... I'd leave it for 6.2 ... maybe mark for removal in 7.0 ?

mauritsvanrees: Agreed.

petschki:
latest plugin release promises to solve our problem with parsing endtime:  https://plugins.jenkins.io/robot/releases/#version_6.2.0 ... 

@gforcada I've updated the plugin, but maybe we need to restart jenkins?

petschki:
ping @gforcada 👋🏼

Is there any chance to debug, why the robotframework plugin still fails to render our reports? https://jenkins.plone.org/job/pull-request-6.2-3.13/428/console 

Since the last update, changelog says this should be fixed https://plugins.jenkins.io/robot/releases/ so I'd like to gather some information for a new issue there. 

When installing the plugin, it said, that jenkins needs a restart. does this happen regularly?

Thank you for your help! 🍀

gforcada:
@petschki sorry, somehow I missed this MR 🫣

I keep updating, every now and then, the server, so it gets restarted frequently.

We are on the latest version, we could create an issue at the plugin to ask for advice.

petschki:
I already did some time ago and reopened it again with these information: https://issues.jenkins.io/browse/JENKINS-75814

petschki:
Looking at the generated XML I'm not sure if the duplicated `<suite id="s2" />` nodes are really correct ... the nested structure looks somehow broken ...

Screenshot 2025-11-18 at 13-16-05

petschki:
Compared to a recent XML output from robotframework 6.0.1

Bildschirmfoto 2025-11-18 um 13 20 19

petschki:
I have also ~created an issue for robotframework here: https://github.com/robotframework/robotframework/issues/5551~ found the root cause for this: https://github.com/collective/robotsuite/pull/34

petschki:
@mauritsvanrees finally we can merge this. I can also backport this to `6.1` what do you think?

mauritsvanrees (approved):

mauritsvanrees:
> @mauritsvanrees finally we can merge this.

Fantastic!  Thanks for your patience and pushing through.

> I can also backport this to `6.1` what do you think?

That should hopefully make the robot tests more stable I guess, so sounds good to me.

PR #1042, opened 2025-10-02
Small update to version pins. [6.1]

mauritsvanrees:
There are more versions we could update, see the [version check GHA](https://github.com/plone/buildout.coredev/actions/runs/18187866827), but I don't want that on the day that I plan to make a Plone bug fix release.  These look safe and wanted.


mauritsvanrees:

PR #1045, opened 2025-10-24
Mxmake update [6.2]

mauritsvanrees:
I ran `mxmake update`, using the latest mxmake version.  See also 
https://github.com/mxstack/mxmake/blob/main/CHANGES.md

This includes a critical change mentioned in that changelog: this no longer runs `mxdev -o`, but `mxdev -f`.
This fixes the following error that I started seeing, [for example on gha](https://github.com/plone/buildout.coredev/actions/runs/18780517910/job/53585253701):

```
RuntimeError: Offline mode:
HTTP reference 'https://zopefoundation.github.io/Zope/releases/5.13/constraints.txt'
not found in cache. Run mxdev in online mode first to populate the cache at .mxdev_cache
```

In recent tests on GHA this week I saw other errors.  [This one](https://github.com/plone/buildout.coredev/actions/runs/18753517860/job/53499289143) seems related to mxmake/mxdev.  [This one](https://github.com/plone/buildout.coredev/actions/runs/18664313285/job/53213014623#step:5:102) not.  Let's see if one of them shows up in this PR.


mauritsvanrees:

jensens (approved):

jensens:
additional, but may be out of scope here, i would set UV_PYTHON.

mauritsvanrees:
> additional, but may be out of scope here, i would set UV_PYTHON.

After merge, the Makefile sets `UV_PYTHON` to `PRIMARY_PYTHON`, which is defined as `python3`.  I think that is better than specifying a specific python version.  At least, this makes the [GHA workflow](https://github.com/plone/buildout.coredev/blob/6.2/.github/workflows/run-make-test.yml) that runs `make test` easier: GHA installs the Python version specified in the matrix, and as a result `python3` should match that version.

Alternatively, I suppose we could switch this GHA to use [`setup-uv`](https://github.com/astral-sh/setup-uv) and pass this the `python-version` from the matrix, which will then set the `UV_PYTHON` env.

jensens:
Yes, that would be the right approach - implemented by me here https://github.com/edutap-eu/edutap.wallet_google/blob/40eb395544d1b3cc51af9e9f52f4c04fad105feb/.github/workflows/tests.yaml#L11-L33 (it uses Tox, but that is interchangeable.

PR #1046, opened 2025-10-24
Mxmake update [6.1]

mauritsvanrees:
Backport of PR #1045 to 6.1.  mxmake/mxdev is less used/tested on 6.1, but we should keep it working.


mauritsvanrees:

jensens (approved):
You could also set `UV_PYTHON?=3.12`  while at it. Or as a separate PR, as you like.

PR #1053, opened 2025-11-21
Use zc.buildout 5.1.0

mauritsvanrees:
And update `pip`, `mr.developer`, and `zc.recipe.egg`.

Buildout now prints a warning at the end:

```
** WARNING **
Some development packages are using old style namespace packages.
You should switch to native namespaces (PEP 420).
If you get a ModuleNotFound or an ImportError when importing a package
from one of these namespaces, you can try this as a temporary workaround:

    pip install horse-with-no-namespace

Note: installing horse-with-no-namespace with buildout will not work.
You must install it into the virtualenv with pip (or uv) before running the buildout.

The following list shows the affected packages and their namespaces:

* plone.restapi: plone
* plone.volto: plone
```

### DIFF
--- a/constraints.txt
+++ b/constraints.txt
@@ -1,16 +1,16 @@
 -c https://zopefoundation.github.io/Zope/releases/5.13/constraints.txt
+horse-with-no-namespace==20251105.1
 packaging==25.0
-pip==25.2
+pip==25.3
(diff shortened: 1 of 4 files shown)


mauritsvanrees:

PR #1054, opened 2025-11-21
Use Zope 6.0b1 and its versions. [6.2]

mauritsvanrees:
This has lots more native namespace packages, and is very close to Python 3.14 support.

PR #1055, opened 2025-11-21
Test Python 3.14.

mauritsvanrees:
This needs two updated versions.  They should work on older Pythons as well.

This builds on the branch of PR #1054.


mauritsvanrees:
Locally it seems to work, but on GHA it quickly fails building lxml 5.4.0.  Maybe update to 6.0.2, which does have wheels for Python 3.14.
Something for later.

mauritsvanrees:
Fox the failures with latest lxml I have opened a [PR in diazo](https://github.com/plone/diazo/pull/92).  But for the moment that PR only updates tox so the tests use `lxml` 6.0.2 and we can see the failures in that repo.

mauritsvanrees:
I close this.  Continued in PR #1070.  After that, we can look at Python 3.14.

PR #1056, opened 2025-11-23
Remove `plone.app.widgets` from `tests.cfg`

petschki:
Since there is no more code needed from `plone.app.widgets` in the core...


petschki:
@mauritsvanrees obviously we could also remove `plone.app.widgets` safely from tests.cfg ... but to be correct, I've created https://github.com/plone/plone.app.widgets/pull/227

petschki:
I'll close this ... lets wait for Plone 7 with removal

PR #1057, opened 2025-11-26
Enable fixed `plone.app.robotframework` tests

petschki:
Tests have been fixed here https://github.com/plone/plone.app.robotframework/pull/178

### DIFF
--- a/tests.cfg
+++ b/tests.cfg
@@ -34,7 +34,7 @@ test-eggs =
     plone.app.redirector
     plone.app.registry
     plone.app.relationfield
-#    plone.app.robotframework [test]
+    plone.app.robotframework [test]
     plone.app.testing [test]
     plone.app.textfield [tests]
     plone.app.theming [test]


petschki:

mauritsvanrees (approved):
All green.  Thanks @1letter and @petschki !

mauritsvanrees:
In commit 8e1abd30c93c54983064ab6e945d63feace3f1fe I added `plone.app.robotframework` to `run-tests.sh` as well for good measure.

mauritsvanrees:
I only notice now that the tests on Jenkins are broken since merging this PR.  That is already 6 days!

See [this error](https://jenkins.plone.org/job/plone-6.2-python-3.13/492/testReport/junit/plone.app.robotframework.tests.test_robot/RobotTestCase/Test_translation/):

> Setup failed:
> Couldn't execute node. Please ensure you have node.js installed and in PATH. See https://nodejs.org/ for instructions. Original error is [Errno 2] No such file or directory: 'node'

I am inclined to revert this PR.  But the strange thing is of course that the build from this PR itself was green.  Same for [this build today](https://jenkins.plone.org/job/pull-request-6.2-3.13/455/) of an unrelated PR.  So it seems there is some difference in setup between the 6.2 PR jobs and the normal 6.2 robot job.  And we can only debug this on the normal job.  So let's keep it for now so we can fix it.

@petschki @1letter Is this something you can look into?

If you don't have time today, then it would be better to revert this change for now.

petschki:
The integration tests from `plone.app.robotframework` strangely try to start tests from the robottest layer, without `rfbrowser init` before. This should not happen. I'll check the testlayers in `plone.app.robotframework` asap.

1letter:
@petschki 
the `REMOTE_LIBRARY_BUNDLE_FIXTURE` is possible wrong in the layer setup of integration tests
```
PLONE_ROBOT_INTEGRATION_TESTING = IntegrationTesting(
    bases=(
        PLONE_ROBOT_FIXTURE,
        REMOTE_LIBRARY_BUNDLE_FIXTURE,
    ),
    name="PloneRobot:Integration",
)
```

petschki:
I tried removing that yesterday, but it did not make any difference. If I add a `ROBOTSUITE_PREFIX`, the robot tests are correctly prefixed and the integration test not, but if you run `bin/test` without `--all` every tests gets started.

@gforcada maybe you have an idea what's wrong here?

1letter:
@petschki `robottestsuite.level = ROBOT_TEST_LEVEL` is the trick, i will fix this in the package

1letter:
@mauritsvanrees i hope this PR helps: https://github.com/plone/plone.app.robotframework/pull/183

mauritsvanrees:
Let me see if I understand:

* The original PR worked because in a PR job we run `rfbrowser init` and `bin/test --all` to run tests at all levels.
* After merge, the robot job worked because it also does `rfbrowser init` and `bin/test --all`, but restricted to the robot test level.
* But the normal core job failed.  This calls `bin/test`, so normally it would not run any robot tests, so there is no need to init the rfbrowser.  But the robot tests in `plone.app.robotframework` were not linked to the `plone.app.testing.ROBOT_TEST_LEVEL`, so they were executed anyway and all failed.

Nice.  Thanks for figuring that out.

PR #1059, opened 2025-11-28
Create target test

wesleybl:
We can use, for example:

`make test package=plone.namedfile`

or

`make test package=plone.namedfile test=test_scaling`

### DIFF
--- a/README-make.md
+++ b/README-make.md
@@ -124,10 +124,24 @@ Run `make zope-start` to start the Zope instance.
 
 ### Run tests
 
-This is not in the Makefile yet.  We need to figure out how best to run the tests first.
-A few sample commands that work for me:
+Use the `make test` target to run unit tests for a package:
 
+```bash
+# Run all unit tests for a package (package is required)
+make test package=plone.namedfile
+
+# Run a specific test
+make test package=plone.namedfile test=test_scaling
 ```
+
+The target automatically detects the correct path for:
+- Packages in development with modern layout (`src/package/src/`)
+- Packages in development with legacy layout (`src/package/`)
+- Packages installed in the virtual environment
+
+You can also run tests directly with `zope-testrunner`:
+
+```bash
 .venv/bin/zope-testrunner --test-path src/Products.CMFPlone -u
 .venv/bin/zope-testrunner --test-path .venv/lib/python3.12/site-packages -s plone.memoize
 ```
--- a/include.mk
+++ b/include.mk
@@ -46,3 +46,32 @@ run-presources: $(PRE_SOURCES_TARGET)
 	@touch $(PRE_SOURCES_TARGET)
 
 INSTALL_TARGETS+=run-presources
+
+##############################################################################
+# test
+##############################################################################
+
+# Usage:
+#   make test package=plone.namedfile
+#   make test package=plone.namedfile test=test_scaling
+#
+.PHONY: test
+test: $(PACKAGES_TARGET)
+ifndef package
+	$(error package is required. Usage: make test package=plone.namedfile [test=test_name])
+endif
+	$(eval TEST_PATH := $(shell \
+		if [ -d "src/$(package)/src" ]; then \
+			echo "src/$(package)/src"; \
+		elif [ -d "src/$(package)" ]; then \
+			echo "src/$(package)"; \
+		else \
+			.venv/bin/python -c "import sysconfig; print(sysconfig.get_paths()['purelib'])"; \
+		fi))
+ifdef test
+	@echo "Running test '$(test)' from package '$(package)' (path: $(TEST_PATH))"
+	@.venv/bin/zope-testrunner -u --test-path $(TEST_PATH) -s $(package) -t $(test)
+else
+	@echo "Running all unit tests from package '$(package)' (path: $(TEST_PATH))"
+	@.venv/bin/zope-testrunner -u --test-path $(TEST_PATH) -s $(package)
+endif


README-make.md
```diff
@@ -124,10 +124,24 @@ Run `make zope-start` to start the Zope instance.
...
-This is not in the Makefile yet.  We need to figure out how best to run the tests first.
-A few sample commands that work for me:
+Use the `make test` target to run unit tests for a package:
 
+```bash
+# Run all unit tests for a package (package is required)
+make test package=plone.namedfile
+
+# Run a specific test
+make test package=plone.namedfile test=test_scaling
```
stevepiercy, line 134: Use generic shell syntax highlighter, don't use comments for instructions because they are low contrast in the rendered code block, and make it easier to copy-paste a single command.

````suggestion
Use the `make test` target to run unit tests for a package.

For example, run all unit tests for a given package.

```shell
make test package=plone.namedfile
```

Run a specific test in the given package.

```shell
make test package=plone.namedfile test=test_scaling
````

File: README-make.md [hunk 2]
```diff
@@ -124,10 +124,24 @@ Run `make zope-start` to start the Zope instance.
...
 ```
+
+The target automatically detects the correct path for:
+- Packages in development with modern layout (`src/package/src/`)
+- Packages in development with legacy layout (`src/package/`)
+- Packages installed in the virtual environment
```
stevepiercy, line 140: No need to capitalize a list item when it is not a sentence.

```suggestion
-   packages in development with modern layout (`src/package/src/`)
-   packages in development with legacy layout (`src/package/`)
-   packages installed in the virtual environment
```

README-make.md
```diff
@@ -124,10 +124,24 @@ Run `make zope-start` to start the Zope instance.
...
+
+You can also run tests directly with `zope-testrunner`:
+
+```bash
```
stevepiercy, line 144:
````suggestion
```shell
````

stevepiercy (changes requested):

README-make.md
wesleybl: @stevepiercy We need to specify somewhere that the `package` parameter is required.

stevepiercy: @wesleybl that's what line 135 does.

wesleybl: Line 135 describes how to run only one test from a given package.

What I'm trying to say is that it's not possible to execute the command:

`make test`

it's necessary to pass the package:

`make test package=plone.namedfile`

See: https://github.com/plone/buildout.coredev/pull/1059/files#diff-35b89ef65700ec4da068cc82250d658677a5aaabec8b91e21d0d3e69987c949fR130

davisagli (commented):
This looks helpful for the use case of testing one package.

The main purpose of buildout.coredev is to provide a place to test all packages together. So it would be nice to also make the `package` parameter optional, and in that case run the tests from all packages. Then we could update the Jenkins builds to use this instead of the buildout.

README-make.md
stevepiercy: I still don't understand what you mean. Line 135 mentions "the given package", and is followed by line 138, which has the `package` parameter.

wesleybl: Line 135 says "the given package," but it doesn't make it clear that this package is required in `make test`. Someone reading this might think it's just an example of running a specific test on a package. And they might try running `make test` thinking it will run all the tests from all the packages. But that won't work.

See: https://github.com/plone/buildout.coredev/pull/1059/files#diff-b2d64a693b39f5304b8f4cfed833e1e499f264e78c5735da98c2d3eb79112777R61

See @davisagli 's https://github.com/plone/buildout.coredev/pull/1059#pullrequestreview-3520219716. It might clarify what I'm trying to say.

wesleybl:
> The main purpose of buildout.coredev is to provide a place to test all packages together. So it would be nice to also make the package parameter optional, and in that case run the tests from all packages

@davisagli The problem is that `site-packages` doesn't only contain Plone packages. Therefore, I can't pass the `site-packages` folder to the `--test-path` of `zope-testrunner`. We would need a list of the packages to be tested, or filter by a prefix.

wesleybl:
@davisagli I see we have this list of packages to be tested for the buildout:

https://github.com/plone/buildout.coredev/blob/6.2/tests.cfg#L6

Can we read from there?

davisagli:
@wesleybl Oh, I see we already have this script to run all the tests (with the same list): https://github.com/plone/buildout.coredev/blob/6.2/run-tests.sh

README-make.md
stevepiercy: @wesleybl let's see whether the package parameter can be optional, and I'll modify this accordingly with a sentence stating whether it's required.

wesleybl: @stevepiercy as @davisagli  said here https://github.com/plone/buildout.coredev/pull/1059#issuecomment-3591002304, we already have a script that runs all the tests. So I think we can leave the `package` parameter required for now.

davisagli: @wesleybl But the run-tests.sh script says it is supposed to be run using `make test`, and that made me realize you are adding a new `test` target even though there is already one: https://github.com/plone/buildout.coredev/blob/6.2/Makefile#L468

wesleybl: Hmm, in version 6.1 it doesn't exist: https://github.com/plone/buildout.coredev/blob/6.1/Makefile

davisagli: I see, it was added in https://github.com/plone/buildout.coredev/pull/1020

It would be nice to have it for both 6.1 and 6.2, with the same implementation in both, and improve it to have a way to optionally run the tests for only one package.

wesleybl: It seems the target already accepts individual packages and tests:

https://github.com/plone/buildout.coredev/blob/6.2/Makefile#L468

I'll test it later.

wesleybl: I checked and actually the target in branch 6.2 doesn't support running only the tests from a specific package.

mauritsvanrees:
I close this in favour of my PR #1071.  That started as only activating the GHA tests on Python 3.14, but now I am trying to run *all* tests, including robot tests, on GHA.  At least for 3.14, as those are not run on Jenkins yet.  For the others I am happy to keep running only the non-layer unit tests.  In the future this may change.

In general, I don't want to put effort into getting a `make test` to work on Plone 6.1.  It is hard enough to get it working on Plone 6.2.

PR #1071, opened 2026-02-18
GHA: test on Python 3.14

mauritsvanrees:


mauritsvanrees:
The unit tests (no layers) were passing on Python 3.14 (and all others).
Now I have `run-tests.sh` to accept `TEST_ARGS` and `TEST_PACKAGE`.  I have adapted the existing jobs to continue to run only unit tests: `make test TEST_ARGS=-u`.  And I have added a workflow to run the full test suite on Python 3.14, divided into robot and non-robot tests.
The robot test job needs some fixes I see, I am looking into that.

mauritsvanrees:
I have removed the new `.github/workflows/run-full-test.yml` for now.

Latest [testall](https://github.com/plone/buildout.coredev/actions/runs/22163580576/job/64085812192) has:

Total: 9459 tests, 28 failures, 8 errors and 37 skipped in 41 minutes 15.422 seconds.

This includes failures in `plone.stringinterp` which work fine for me locally.  But also a few `CMFPlone` tests that need minor fixes for 3.14.  And some failures because i18n does not seem to work.

And latest [testrobot](https://github.com/plone/buildout.coredev/actions/runs/22163580576/job/64085812165) has:

Total: 184 tests, 3 failures, 0 errors and 0 skipped in 10 minutes 7.599 seconds.

This also looks like i18n problems.

Ah, with buildout we set some env vars in `bin/test` which we may need to take over, which may help for i18n, and possibly also for speed:

```
os.environ['CHAMELEON_CACHE'] = '/Users/maurits/community/plone-coredev/6.2/var/cache'
os.environ['ROBOTSUITE_LOGLEVEL'] = 'ERROR'
os.environ['ROBOTSUITE_APPEND_OUTPUT_XML'] = '1'
os.environ['zope_i18n_compile_mo_files'] = 'true'
```

Anyway, as said, I have removed the full test action for now.  It shows that the approach with passing `TEST_ARGS` and `TEST_PACKAGE` works.  I will redo that later and first try it with Python 3.13, where we know the tests should pass.

The main thing is that we call `make test TEST_ARGS=-u` and that this works for Python 3.14 as well, on both Ubuntu and Mac.  And the buildout finishes successfully or Python 3.14 as well, on Ubuntu, Mac, and Windows.

Tests are green now.  I will squash and merge.

PR #1072, opened 2026-02-19
Add workflow to run full test suite on GHA.

mauritsvanrees:
This is a continuation from [PR 1071](https://github.com/plone/buildout.coredev/pull/1071). Especially see [this comment](https://github.com/plone/buildout.coredev/pull/1071#issuecomment-3926570133) on what did not yet work at that time.

As a start, we use Python 3.13.  Later we can move this to 3.14.  Then we have this as long as 3.14 does not work on Jenkins. For Jenkins, see https://github.com/plone/jenkins.plone.org/issues/384.

### DIFF
--- a/.github/workflows/full-shared.yml
+++ b/.github/workflows/full-shared.yml
@@ -1,0 +1,59 @@
+name: "Shared setup for running full tests"
+
+on:
+  workflow_call:
+    inputs:
+      python-version:
+        required: false
+        type: string
+        default: "3.14"
+      prepare-command:
+        required: false
+        type: string
+        default: ""
+      test-command:
+        required: false
+        type: string
+        default: "make test"
+
+jobs:
+  test:
+    name: test
+    runs-on: ubuntu-latest
+    env:
+        PYTHONWARNINGS: ignore
+        GIT_CLONE_DEPTH: 1
+        CHAMELEON_CACHE: 'chameleon_cache'
+        zope_i18n_compile_mo_files: 'true'
+        ROBOTSUITE_LOGLEVEL: ERROR
+        ROBOTSUITE_PREFIX: ONLYROBOT
+    steps:
+    - name: OS packages
+      run:
+        sudo apt-get update && sudo apt-get install -y wv poppler-utils pdftohtml gettext
+    - name: locale
+      # needed for CMFPlone testUnicodeSplitter test on Ubuntu
+      run: |
+        sudo locale-gen nl_NL@euro
+        sudo update-locale
+    - uses: actions/checkout@v4
+    - name: Set up Python ${{ inputs.python-version}}
+      uses: actions/setup-python@v5
+      with:
+        python-version: ${{ inputs.python-version}}
+        cache: 'pip'
+    - name: Create Chameleon cache dir
+      run: |
+        mkdir -p "${CHAMELEON_CACHE}"
+    - name: Install
+      # Run 'make install' twice. Usually the second run still does something.
+      run: |
+        make install
+        make install
+    - name: Prepare
+      if: ${{ inputs.prepare-command }}
+      run: |
+        ${{ inputs.prepare-command }}
+    - name: Test
+      run: |
+        ${{ inputs.test-command }}
--- a/.github/workflows/run-full-test.yml
+++ b/.github/workflows/run-full-test.yml
@@ -1,5 +1,10 @@
 name: Run full test suite
-on: push
+on:
+  push:
+    branches:
+      - "6.2"
+  pull_request:
+  workflow_dispatch:
 jobs:
   testfunctional:
     name: ${{ matrix.python-version }} ${{ matrix.module }} functional
--- a/.github/workflows/run-full-test.yml
+++ b/.github/workflows/run-full-test.yml
@@ -1,0 +1,39 @@
+name: Run full test suite
+on: push
+jobs:
+  testfunctional:
+    name: ${{ matrix.python-version }} ${{ matrix.module }} functional
+    strategy:
+      matrix:
+        python-version:
+        - "3.13"
+        - "3.14"
+        module:
+        - Products.CMFPlone
+        - plone.app
+        - "!Products -m !plone.app"
+    uses: ./.github/workflows/full-shared.yml
+    with:
+      python-version: ${{ matrix.python-version }}
+      # Note: the ubuntu-latest runners have 4 cpus, so we could run tests in
+      # parallel by adding '-j 4' to the TEST_ARGS.  But this can lead to a
+      # strange error in zope.i18n.zcml.registerTranslations:
+      # struct.error: unpack requires a buffer of 4 bytes
+      test-command: make test TEST_ARGS='-f -m ${{ matrix.module }}'
+
+  testrobot:
+    name: ${{ matrix.python-version }} ${{ matrix.module }} robots
+    strategy:
+      matrix:
+        python-version:
+        - "3.13"
+        - "3.14"
+        module:
+        - Products.CMFPlone
+        - "!Products.CMFPlone"
+    uses: ./.github/workflows/full-shared.yml
+    with:
+      python-version: ${{ matrix.python-version }}
+      prepare-command: make rfbrowser
+      # Note: robot tests should not be run in parallel.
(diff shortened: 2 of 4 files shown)


mauritsvanrees:
Yes, it works on Python 3.13:

* Running [all tests except robot](https://github.com/plone/buildout.coredev/actions/runs/22180002689/job/64138544831?pr=1072): "Total: 9463 tests, 0 failures, 0 errors and 37 skipped in 38 minutes 40.792 seconds."
* Running [robot tests](https://github.com/plone/buildout.coredev/actions/runs/22180002689/job/64138544850?pr=1072): "Total: 184 tests, 0 failures, 0 errors and 0 skipped in 9 minutes 9.142 seconds."

Those are the same test numbers as I see on Jenkins.  On Jenkins the robot tests are a few seconds slower, and the other tests are one minute faster.  That difference is negligible.

Now we could try it on 3.14.  This needs at least [one CMFPlone PR](https://github.com/plone/Products.CMFPlone/pull/4280) to be merged first.  Maybe more.

And then we can see if we want to keep it like this.  It does not seem so attractive (and a bit wasteful) to have to wait for a job that takes 38 minutes.  We could change that one to use `TEST_PACKAGE=Products.CMFPlone`.  At least for true coredev PRs that would be better: if I change a few version pins I am already happy if the buildout finishes correctly, especially if these are versions from `versions-extra.cfg`.  When a commit to 6.2 is done, we could still run all tests, at least as long as 3.14 does not work on Jenkins.

For now I marked this PR as draft.

mauritsvanrees:
Alright!   All tests on Python 3.14 pass:

* [testall job](https://github.com/plone/buildout.coredev/actions/runs/22312952880/job/64549786413#step:7:1286): Total: 9485 tests, 0 failures, 0 errors and 40 skipped in 39 minutes 34.284 seconds.
* [testrobot job](https://github.com/plone/buildout.coredev/actions/runs/22312952880/job/64549786231#step:7:120): Total: 184 tests, 0 failures, 0 errors and 0 skipped in 9 minutes 22.097 seconds.

davisagli:
@mauritsvanrees Nice work!

gforcada:
Oh wow, that's very nice! 🌟 

Do we have unlimited runner minutes? 🤔

mauritsvanrees:
> Oh wow, that's very nice! 🌟

Thanks.
 
> Do we have unlimited runner minutes? 🤔

Should be unlimited, as this is a public repository:
https://docs.github.com/en/billing/concepts/product-billing/github-actions

davisagli:
Do the runners have multiple CPUs? Can we speed things up by running multiple layers in parallel with the `-j` option?

mauritsvanrees:
I am testing some stuff on a separate branch [`maurits-fun-full-tests-on-gha-shared`](https://github.com/plone/buildout.coredev/tree/refs/heads/maurits-fun-full-tests-on-gha-shared) (no PR), with all other workflows removed.  I have split the full tests up into [four jobs](https://github.com/plone/buildout.coredev/blob/refs/heads/maurits-fun-full-tests-on-gha-shared/.github/workflows/run-full-test.yml), like this:

* functional Products tests (about 2600)
* functional plone.app tests (about 2300)
* functional other tests (about 2400)
* robot tests (184 tests)
* no pure unit tests, as they are handled already in the other workflows

Here is the [latest test run](https://github.com/plone/buildout.coredev/actions/runs/22356347508).
This shows that the `plone.app` tests take half an hour, which I did not expect.
I am experimenting a bit locally to see if I can make a more balanced split.

> Do the runners have multiple CPUs? Can we speed things up by running multiple layers in parallel with the `-j` option?

The [docs say](https://docs.github.com/en/actions/reference/runners/github-hosted-runners) that `ubuntu-latest` which we use has 4 CPUs.  So that is worth a try.

gforcada:
I see that `plone.app.layout` also takes quite some time:

```
Running plone.app.layout.testing.plone.app.layout:Integration tests:
  Tear down plone.app.layout.testing.plone.app.layout:Functional in 0.000 seconds.
  Set up plone.app.layout.testing.plone.app.layout:Integration in 0.000 seconds.
  Ran 139 tests with 0 failures, 0 errors and 0 skipped in 9 minutes 20.090 seconds.
```

mauritsvanrees:
Yes, I noticed `plone.app.layout` as well.  Locally `test_first_page_has_batch_size_count_items` takes 1 minute 40 seconds.

mauritsvanrees:
After our fixes to multilingual and layout, the `plone.app` tests on my other branch are [much faster](https://github.com/plone/buildout.coredev/actions/runs/22388583602), going from 30 minutes down to almost 7:

Screenshot 2026-02-25 at 10 39 11

Now I am trying out running the tests in parallel with 4 threads, and two extra jobs.  See [this run](https://github.com/plone/buildout.coredev/actions/runs/22389533090):

Screenshot 2026-02-25 at 10 39 37

This does not have that much of an effect.  The Products tests go from 5 and a half to 4 minutes, but the rest are more or less the same.  The Robot tests without parallelisation were also a bit faster this time, so there is some random influence as well.  I suppose even small gains are nice here, so we can keep it.

I tried the Robot tests in parallel as well.  I already thought this may not work, and indeed it does not.  The errors are a bit weird though, with `gettext` not able to [parse a translation file](https://github.com/plone/buildout.coredev/actions/runs/22389533090/job/64807731489#step:9:84).  Anyway, let's not run the robot tests in parallel.

For comparison I added a job for running all tests (except robot) in parallel, without the packages split up, and that takes 15 minutes.  Even after the speed up of multilingual/layout, this is still long.  So the split is good.

Let me try a bit more and then update the current branch.

mauritsvanrees:
This is working now.  I have added a matrix and am running tests on 3.13 and 3.14.  I have added documentation in `README-make.md` with lots of examples, taking some hints from what @wesleybl wrote in his PR #1059.  I have rebased and force pushed to make the history cleaner.

Ready for review, and can be merged.

.github/workflows/full-shared.yml
```diff
@@ -0,0 +1,59 @@
...
+        python-version: ${{ inputs.python-version}}
+        cache: 'pip'
+    - name: Create Chameleon cache dir
+      run: |
+        mkdir -p "${CHAMELEON_CACHE}"
+    - name: Install
+      # Run 'make install' twice. Usually the second run still does something.
+      run: |
+        make install
+        make install
```
gforcada, line 52: twice on purpose? 🤔

.github/workflows/run-full-test.yml
```diff
@@ -0,0 +1,44 @@
...
+    name: ${{ matrix.python-version }} ${{ matrix.module }} functional
+    strategy:
+      matrix:
+        python-version:
+        - "3.13"
+        - "3.14"
+        module:
+        - Products.CMFPlone
+        - plone.app
+        - "!Products -m !plone.app"
```
gforcada, line 19: The only `Products` tests that we have are on CMFPlone only? Otherwise we would be skipping them, or? 🤔

gforcada (approved):
Very nice! 🌟 

I only have two minor questions 👍🏾

.github/workflows/full-shared.yml
mauritsvanrees: Yes, on purpose.  If you `make clean` locally and then `make install`, then the next `make install` still usually does something.  Ideally this should not be the case: some improvement to the mxmake setup may be possible.  But that is outside the range of this PR.

.github/workflows/run-full-test.yml
mauritsvanrees: Oops, my bad.  That was meant to be the same in both cases.
Fixed in bd631a239e8f43425c1c66712dc775fa064742b6 by having `Products` in both cases.  Picking `Products.CMFPlone` would be fine as well.

.github/workflows/full-shared.yml
wesleybl: It's best to leave a comment, because other people might think it's a mistake.

mauritsvanrees: There already is a comment:

```
# Run 'make install' twice. Usually the second run still does something.
```

gforcada (approved):

PR #1075, opened 2026-02-26
Update package webresource to 2.0.0

1letter:
[webresource](https://pypi.org/project/webresource/) is released 2.0.0, check what happen


mauritsvanrees (approved):
That seems to work.

I am using this PR to check if the newly defined 6.2-3.14 PR job on Jenkins works.  Here is the [console](https://jenkins.plone.org/view/all/job/pull-request-6.2-3.14/1/console).  If that is green, this can be merged.  If not green, it could also be a problem with the job.

PR #1081, opened 2026-03-25
Install playwright dependencies to fix connection errors

petschki:
try to fix occasional connection errors by installing playwright system dependencies...

### DIFF
--- a/include.mk
+++ b/include.mk
@@ -52,6 +52,6 @@ INSTALL_TARGETS+=run-presources
 .PHONY: rfbrowser
 rfbrowser: $(FILES_TARGET) $(SOURCES_TARGET) $(PACKAGES_TARGET) $(TEST_TARGET) $(TEST_DEPENDENCY_TARGETS)
 	@echo "Installing Playwright system dependencies"
-	@npx playwright install-deps
+	@npx playwright install-deps chromium
 	@echo "Initializing robotframework browser"
 	@/usr/bin/env bash -c "rfbrowser init chromium"
--- a/include.mk
+++ b/include.mk
@@ -51,5 +51,7 @@ INSTALL_TARGETS+=run-presources
 
 .PHONY: rfbrowser
 rfbrowser: $(FILES_TARGET) $(SOURCES_TARGET) $(PACKAGES_TARGET) $(TEST_TARGET) $(TEST_DEPENDENCY_TARGETS)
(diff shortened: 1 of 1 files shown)


mauritsvanrees (approved):
I don't know what this exactly does, and how it would help, but it sounds okay.

Locally I need to press Enter, otherwise nothing happens:

```
$ make rfbrowser 
Installing Playwright system dependencies
Need to install the following packages:
playwright@1.58.2
Ok to proceed? (y) 
```

But on GHA it continues, so should be okay.  And I don't see a command line option to influence this.

If we are targeting chromium, we can explicitly call `npx playwright install-deps chromium`

We should do the same on `plone.meta`, right?

mauritsvanrees:
Might help on Jenkins as well, if we update those scripts.

Would running `npx playwright install-deps chromium` once on each Jenkins node already help?

petschki:
I've tested this on jenkins and added it here https://github.com/plone/jenkins.plone.org/pull/386 but on jenkins, we need sudo rights to install playwright dependencies, so not sure how to proceed there... also the `rfbrowser init` on jenkins doesn't show warnings about missing systemdependencies ... GHA shows it here for example https://github.com/plone/buildout.coredev/actions/runs/23535806334/job/68510562717#step:8:54

petschki:
> Would running `npx playwright install-deps chromium` once on each Jenkins node already help?

yes, as a sudo user this is enough once ... so we could add it to `plone.jenkins_node` eventually?

include.mk
```diff
@@ -51,5 +51,7 @@ INSTALL_TARGETS+=run-presources
 
 .PHONY: rfbrowser
 rfbrowser: $(FILES_TARGET) $(SOURCES_TARGET) $(PACKAGES_TARGET) $(TEST_TARGET) $(TEST_DEPENDENCY_TARGETS)
+	@echo "Installing Playwright system dependencies"
+	@npx playwright install-deps
```
mauritsvanrees:
```suggestion
	@npx playwright install-deps chromium
```

mauritsvanrees:
@petschki Should this PR be closed?  We closed the [related one](https://github.com/plone/jenkins.plone.org/pull/386) for Jenkins.

petschki:
I'm not sure...  currently tests are running stable. So lets do not disturb them again and close this 😉

PR #1084, opened 2026-03-31
Add --path for local src/*/src in run-tests.sh to prefer local checkouts

### LINKED ISSUE #1083: `make test TEST_PACKAGE=plone.restapi` does not use the package in the src folder.
I ran `make test TEST_PACKAGE=plone.restapi` and `plone.restapi` was in the checkout in the `src` folder. But the code from the src folder wasn't used; instead, it was the code from the `site-packages` folder.

wesleybl:
Closes #1083

### DIFF
--- a/run-tests.sh
+++ b/run-tests.sh
@@ -160,8 +160,15 @@ else
     TEST_ARGS="$TEST_ARGS -s $TEST_PACKAGE"
 fi
 
+# Add --path for each package checked out in the src/ folder so zope-testrunner
+# uses the local source instead of the installed site-packages copy.
+PATH_ARGS=""
+for src_pkg in src/*/src; do
+    PATH_ARGS="$PATH_ARGS --path $src_pkg"
+done
+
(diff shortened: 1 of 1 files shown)


mauritsvanrees:
Locally I have 217 packages in my `src` directory.  With this change, the script would take those tests instead of from the packages that are actually installed in the virtualenv.  That is not what we want.  Some of these directories may be very outdated, making the tests fail.

The intention of the script is to run the tests of the installed packages.  If a checkout is not active, we should not take the tests from there.  It's the same as the `bin/test` script that we create when using buildout: you can run `bin/test -s plone.restapi` and currently it will run the tests that it finds in site-packages, because the checkout is not active.

Note that this script is using the `--auto-path` option that I [added last year](https://github.com/zopefoundation/zope.testrunner/pull/198) to `zope.testrunner`.  This makes sure that when you test a specific package, the installed path of that package is automatically added to the paths where tests are searched.

PR #1085, opened 2026-04-01
Add 'make manage' target, update readme.

mauritsvanrees:
See https://github.com/plone/buildout.coredev/issues/1083

### DIFF
--- a/README-make.md
+++ b/README-make.md
@@ -106,7 +106,7 @@ These large Makefiles are tricky.
 
 Run `make zope-start` to start the Zope instance.
 
-### Manage checkouts and versions
+### The `manage` command
 
 The `manage` command from `plone.releaser` can be used in various ways.
 It knows how to query or update files from both buildout and pip:
@@ -117,17 +117,37 @@ See the help text:
 make manage ARGS="--help"
 ```
 
+### Manage checkouts
+
 Add a package to the checkouts:
 
 ```shell
 make manage ARGS="add-checkout plone.restapi"
 ```
 
+This does not yet checkout the package: it only updates the `checkouts.cfg` and `mxcheckouts.ini` files.
+To *use* the new checkout, you still need to run `make install`.
+If the new checkout is still not used, you can more agressively get a new environment:
+`make mxenv-clean` and `make install`.
+
+Remove a package from the checkouts:
+
+```shell
+make manage ARGS="remove-checkout plone.restapi"
+```
+
+Similarly, you still need to call `make install` after this.
+
+### Manage versions
+
 Update a package version:
 
 ```shell
 make manage ARGS="set-package-version plone.restapi 10.0.0"
 ```
+
+Again, you still need to call `make install` to actually install the new version.
+
 
 ### Run tests
 
@@ -157,6 +177,18 @@ make test
 Note that this tests the *installed* version of the package.
 This can be a final version in `site-packages` or a source checkout.
 See the section "Manage checkouts and versions" on how to influence what is installed.
+
+Note: Simply having a package directory under `src/` is not enough for `make test` to pick it up.
+This may be an old checkout that is not currently active.
+You need to explicitly add it as a checkout and reinstall:
+
+```
+make manage ARGS="add-checkout plone.restapi"
+make mxenv-clean
+make install
+```
+
+Only after this will `make test TEST_PACKAGE=plone.restapi` use the source from src/.
 
 Under the hood, the `test` target calls `zope-testrunner`.
 To find out which arguments this supports, ask for help:
--- a/README-make.md
+++ b/README-make.md
@@ -99,11 +99,35 @@ Main results:
 * In `.venv` there is a virtual environment with Plone and all test requirements installed.
 * In `instance` a Zope instance is created.
 
+Important: you may need to run `make install` twice, especially after you change something.
+These large Makefiles are tricky.
 
 ### Start
 
 Run `make zope-start` to start the Zope instance.
 
+### Manage checkouts and versions
+
+The `manage` command from `plone.releaser` can be used in various ways.
+It knows how to query or update files from both buildout and pip:
+
+See the help text:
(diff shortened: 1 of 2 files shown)


README-make.md
```diff
@@ -99,11 +99,35 @@ Main results:
...
+See the help text:
+
+```shell
+make manage ARGS="--help"
+```
+
+Add a package to the checkouts:
+
+```shell
+make manage ARGS="add-checkout plone.restapi"
```
wesleybl, line 122: I had to run `make mxenv-clean` and then `make install` for the checkout to work. I think it's worth mentioning.

README-make.md
```diff
@@ -99,11 +99,35 @@ Main results:
...
 * In `instance` a Zope instance is created.
 
+Important: you may need to run `make install` twice, especially after you change something.
+These large Makefiles are tricky.
 
 ### Start
 
 Run `make zope-start` to start the Zope instance.
 
+### Manage checkouts and versions
```
wesleybl: I think we can be more explicit about the tests. I think we could add something like:

```
Note: Simply having a package directory under src/ is not enough for make test to pick it up.
You need to explicitly add it as a checkout and reinstall:

make manage ARGS="add-checkout plone.restapi"
make mxenv-clean
make install

Only after this will `make test TEST_PACKAGE=plone.restapi` use the source from src/.
```

wesleybl (changes requested):

mauritsvanrees:
@wesleybl I adapted and included your comments.  Can you review again please?

README-make.md
```diff
@@ -99,11 +99,55 @@ Main results:
...
+### Manage checkouts
+
+Add a package to the checkouts:
+
+```shell
+make manage ARGS="add-checkout plone.restapi"
+```
+
+This does not yet checkout the package: it only updates the `checkouts.cfg` and `mxcheckouts.ini` files.
+To *use* the new checkout, you still need to run `make install`.
```
wesleybl, line 126: If the package doesn't exist in `src` and I do:

`make manage ARGS="add-checkout plone.restapi"`

executing only `make install`, the clone doesn't occur. The `mxdev` command is only executed if `mx.ini` is modified. But `make manage` modifies `mxcheckouts.ini` and not `mx.ini`. See the rule:

https://github.com/plone/buildout.coredev/blob/bd4604c6d77b0ed453f1d0dac0afa9dc2ba3c1de/Makefile#L336-L340

We must also add the other files that, if modified, should execute `mxdev`. We must do this in `include.mk`:

```bash
$(SOURCES_TARGET): mxcheckouts.ini mxsources.ini mxtests.ini
```

README-make.md
```diff
@@ -130,7 +174,23 @@ export TEST_PACKAGE=plone.namedfile
...
+This can be a final version in `site-packages` or a source checkout.
+See the section "Manage checkouts and versions" on how to influence what is installed.
+
+Note: Simply having a package directory under `src/` is not enough for `make test` to pick it up.
+This may be an old checkout that is not currently active.
+You need to explicitly add it as a checkout and reinstall:
+
+```
+make manage ARGS="add-checkout plone.restapi"
+make mxenv-clean
```
wesleybl: `make mxenv-clean` that's no longer necessary with my previous comment.

README-make.md
```diff
@@ -99,11 +99,55 @@ Main results:
 * In `.venv` there is a virtual environment with Plone and all test requirements installed.
 * In `instance` a Zope instance is created.
 
+Important: you may need to run `make install` twice, especially after you change something.
```
wesleybl: Perhaps it's no longer necessary to run `make install` twice, after my suggestion.

wesleybl (changes requested):

README-make.md
mauritsvanrees: That seems to work, at least for this case, which is good!

But I am not sure.  You basically then have this:

```
SOURCES_TARGET:=$(SENTINEL_FOLDER)/sources.sentinel 
$(SOURCES_TARGET): $(PROJECT_CONFIG) $(MXENV_TARGET) 
	@echo "Checkout project sources" 
	@mxdev -f -c $(PROJECT_CONFIG) 
	@touch $(SOURCES_TARGET)
$(SOURCES_TARGET): mxcheckouts.ini mxsources.ini mxtests.ini
```

It seems to me that this line overrides the second line, so maybe they need to be combined.  Or does this happen automatically?  The commands are still executed, so that part works.

@jensens Can you advise here?

Maybe in `mxmake` we could define it like this (if I spell it correctly):

```
SOURCES_TARGETS=$(PROJECT_CONFIG) $(MXENV_TARGET)
$(SOURCES_TARGET): $(SOURCES_TARGETS)
    ...
```

And then in our `include.mk` we could add:

```
SOURCES_TARGETS+=mxcheckouts.ini mxsources.ini mxtests.ini
```

But I guess that line is included too late, so it would have no effect.

wesleybl: >The commands are still executed, so that part works.

Yes, it works. make merges all prerequisites into a single list. You can leave a comment to clarify this, since it generated some confusion.

But your idea of ​​`SOURCES_TARGETS+=` is good. It makes things clearer. But it would require a change in `mxmake`.

So my recommendation is to add the comment and open an issue in `mxmake`.

mauritsvanrees: I have added your line and a comment.

I created https://github.com/mxstack/mxmake/issues/72

README-make.md
mauritsvanrees: I have removed that line.

README-make.md
mauritsvanrees: I have removed this comment.

mauritsvanrees:
I have finally added your comment:

```
$(SOURCES_TARGET): mxcheckouts.ini mxsources.ini mxtests.ini
```

I have rebased on 6.2.  All seems fine now, so I merge.

PR #1090, opened 2026-05-20
DO NOT MERGE: run 'manage pip2buildout'

mauritsvanrees:
The first commits are small manual edits to prepare.  And the last commit calls `manage pip2buildout` as a test for https://github.com/plone/plone.releaser/pull/99.  So see the last commit for the actual changes that are done by this command.

Later we can decide if we really do this.  It would be for Plone 6.3.

Note that comments are removed.  Some of those should then be manually moved from buildout files to pip/mxdev files.


mauritsvanrees:
Let's check Jenkins for good measure.

PR #1092, opened 2026-05-21
DO NOT MERGE: start 6.3 branch

mauritsvanrees:
This should not be merged.  The 6.3 branch can live on its own.
But it is good if someone looks at the initial changes, and a PR is useful for that.


gforcada (approved):

davisagli:
@mauritsvanrees I added branch protection for the 6.2 and 6.3 branches

mauritsvanrees:
I have cleaned up some more, for example removing no longer needed buildout parts, and untested ecosystem packages.
I am done for now.
Maybe another check by someone, and then we can just close this PR, and keep the branch.

mauritsvanrees:
This no longer needs to be a PR.  I close this.  6.3 will be a normal branch now, just like the others.

PR #1094, opened 2026-05-28
add plone.app.i18n test dependencies

petschki:
And pin `z3c.jbot` version.


davisagli (approved):

mauritsvanrees (approved):
I have added two commits, to update `constraints.txt` as well, and to remove the pin from the ecosystem versions.

Can be merged when green.  (I think it is okay to not ask Jenkins.)

